PR #14, opened 2021-04-08
Fix readthedocs generation

innerlee:


requirements/readthedocs.txt
```diff
@@ -1,3 +1,18 @@
...
+Pillow<=6.2.2
+Polygon3
+pyclipper
+pycocotools
+python-Levenshtein
+scikit-image
+scipy
+shapely
+skimage
+titlecase
```
hellock, line 16: There is no need to install all dependencies in readthedocs. Most ones can be mocked.

cuhk-hbsun (approved):
pass

PR #24, opened 2021-04-08
Replace python-Levenshtein with RapidFuzz

maxbachmann:
This PR replaces python-Levenshtein with RapidFuzz (disclaimer: I am the author) for the following reasons:
- python-Levenshtein is the only GPL licensed dependency (RapidFuzz is MIT licensed)
- RapidFuzz provides wheels, while this currently requires workarounds like the `Levenshtein` library for python-Levenshtein (This will probably get better once a new maintainer for python-Levenshtein is found)
- The Levenshtein implementation in RapidFuzz is a lot faster:

### DIFF
--- a/mmocr/core/evaluation/ocr_metric.py
+++ b/mmocr/core/evaluation/ocr_metric.py
@@ -1,7 +1,7 @@
 import re
 from difflib import SequenceMatcher
 
-import Levenshtein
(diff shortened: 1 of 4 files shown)


innerlee:
Thanks for the proposal, we will evaluate it :D

cuhk-hbsun (approved):

PR #38, opened 2021-04-09
rename datasets.md

cuhk-hbsun:


jeffreykuang (commented):
why rename datasets.md?

innerlee:
Yeah if we can fix the readthedocs without renaming this, then we do not rename it :)

jeffreykuang (approved):

PR #86, opened 2021-04-18
Add support for batch inference

sbugallo:
* Add support for batch inference
* Add batch inference demo script


cuhk-hbsun:
Thanks for your contribution, but there are some problems to use batch-mode:
1. Most of the config cannot use batch mode during test infernece, because of `aug_test`. For example, `MultiScaleFlipAug` in configs/textdet/psenet/psenet_r50_fpnf_600e_ctw1500.py, `MultiRotateAugOCR` in configs/textrecog/sar/sar_r31_parallel_decoder_academic.py
2. Currently, we use `tools/test_imgs.py` (text detection) and `tools/ocr_test_imgs.py` (text recognize) for the visualization of image list, while in `not-batch mode`.

sbugallo:
Thank you for your feedback. Both scripts you mentioned (`tools/test_imgs.py` and `tools/ocr_test_imgs.py`) are perfectly fine except for cases where you have time restrictions (e.g., real time pipelines). This is be very noticeable when using text detection followed by text recognition. 

I understand that the incompatibility of some configs are a major problem. Maybe we can find a workaround to at least enable this feature with compatible configs. I would gladly help making `aug_test` compatible if needed.

cuhk-hbsun:
Thanks a lot, and hope to improve it with U.

sbugallo:
Hi @cuhk-hbsun . I did a few updates. Please, have a look at then if you can:

* I changed [these lines](https://github.com/open-mmlab/mmocr/pull/86/files#diff-5397eb1f81de52957191f6910a42d74c1520ca95b6b7838baa1e84047696dc70R59) so both detectors and recognizers receive the same data types and formats when doing inference.
* [This change](https://github.com/open-mmlab/mmocr/pull/86/files#diff-70a0b1818037e4e6cc2aaa90b78b55c4d6e95bcb2736f01e89f2937a1acf5f3aR40) may be a bit naive but it enables batch mode in single-stage detectors.
* I moved [this logic](https://github.com/open-mmlab/mmocr/pull/86/files#diff-d5f388d02f23632af93e57187608e1e11d3532ad67bc6065c9baaa90253c395dR99) inside the base text recognizer to make the inference function a little more general (this is related to the first point).

The `test_model_inference.py` file is still a work in progress but I prefer to check if the previous changes look good enough for you before changing more things (and I kind of debug errors easier this way :) ).

Please note that pytest is failing due to this file eating a lot of resources and thus the process is killed. Tests are passing, though.

antoniolanza1996:
Hey @sbugallo thanks for implementing this feature, I'm also interested on it.

Have you noticed a better inference speed with your batch inference implementation?

sbugallo:
Hi @antoniolanza1996.  I'm using the NRTR R31-1/8-1/4 model to do text recognition on 5 image regions. Running the pipeline sequentially (that is running 5 inferences) gives me a maximum speed of ~4 FPS on a GTX 1070. Using batch inference to process the everything in a single iteration speeds the pipeline to ~22 FPS, which is quite expected (around five times faster).

cuhk-hbsun (approved):

PR #130, opened 2021-04-27
Fix issue 122

### LINKED ISSUE #122: BUG report: points can be duplicated due to transform.
Here https://github.com/open-mmlab/mmocr/blob/1cd3c8d3e202d2dd436daa550b32bf35dcbba803/mmocr/datasets/pipelines/textdet_targets/textsnake_targets.py#L347 , the polygon after transformation will be float type, when you change the points from float to int, some points will be the same. Like the screen shot below.
image
fix it by
```python
polygon_points = np.array(
                list({
                    tuple(np.array(p, dtype=np.int32))
                    for p in text_instance
                })).reshape(-1, 2)
```

HolyCrap96:
Fix #122: adapt textsnake targets to other datasets

### DIFF
--- a/mmocr/datasets/pipelines/transforms.py
+++ b/mmocr/datasets/pipelines/transforms.py
@@ -409,8 +409,12 @@ class RandomCropPolyInstances:
         return start, end
 
     def sample_crop_box(self, img_size, results):
-        """Generate crop box and make sure not to crop the polygon
-        instances."""
+        """Generate crop box and make sure not to crop the polygon instances.
+
+        Args:
+            img_size (tuple(int)): The image size (h, w).
+            results (dict): The results dict.
+        """
 
         assert isinstance(img_size, tuple)
         h, w = img_size[:2]
--- a/mmocr/datasets/pipelines/transforms.py
+++ b/mmocr/datasets/pipelines/transforms.py
@@ -408,37 +408,38 @@ class RandomCropPolyInstances:
                                 region_ends[region_ind])
         return start, end
 
-    def sample_crop_box(self, img_size, masks):
-        """Generate crop box and make sure not to crop the polygon instances.
-
-        Args:
-            img_size (tuple(int)): The image size.
-            masks (list[list[ndarray]]): The polygon masks.
-        """
+    def sample_crop_box(self, img_size, results):
+        """Generate crop box and make sure not to crop the polygon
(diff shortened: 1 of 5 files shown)


mmocr/datasets/pipelines/transforms.py
```diff
@@ -408,37 +408,38 @@ def sample_valid_start_end(self, valid_array, min_len, max_start, min_end):
...
-    def sample_crop_box(self, img_size, masks):
-        """Generate crop box and make sure not to crop the polygon instances.
-
-        Args:
-            img_size (tuple(int)): The image size.
-            masks (list[list[ndarray]]): The polygon masks.
-        """
+    def sample_crop_box(self, img_size, results):
+        """Generate crop box and make sure not to crop the polygon
+        instances."""
```
jeffreykuang, line 413: docstring for args

mmocr/datasets/pipelines/transforms.py
```diff
@@ -408,37 +408,38 @@ def sample_valid_start_end(self, valid_array, min_len, max_start, min_end):
...
-        for mask in masks:
-            assert len(mask) == 1
-            mask = mask[0].reshape((-1, 2)).astype(np.int32)
-            clip_x = np.clip(mask[:, 0], 0, w - 1)
-            clip_y = np.clip(mask[:, 1], 0, h - 1)
-            min_x, max_x = np.min(clip_x), np.max(clip_x)
-            min_y, max_y = np.min(clip_y), np.max(clip_y)
+        for key in results.get('mask_fields', []):
+            if len(results[key].masks) == 0:
+                continue
```
jeffreykuang, line 431: can you add unit test to cover 431?

mmocr/datasets/pipelines/transforms.py
```diff
@@ -458,9 +459,10 @@ def crop_img(self, img, bbox):
         return img[bbox[1]:bbox[3], bbox[0]:bbox[2]]
 
     def __call__(self, results):
+        if len(results[self.instance_key].masks) < 1:
+            return results
```
jeffreykuang, line 463: can you add unittest to cover l463?

HolyCrap96: unit test added

jeffreykuang (approved):

PR #135, opened 2021-04-28
refactor tps config

cuhk-hbsun:


configs/textrecog/tps/crnn_tps_academic_dataset.py
```diff
@@ -26,7 +26,7 @@
         img_size=(32, 100),
         rectified_img_size=(32, 100),
         num_img_channel=1),
-    backbone=dict(type='VeryDeepVgg', leakyRelu=False, input_channels=1),
+    backbone=dict(type='VeryDeepVgg', leaky_relu=False, input_channels=1),
```
innerlee, line 29: So here is a missing item in the previous refactor. Are there any more items missed?

innerlee (approved):

configs/textrecog/tps/crnn_tps_academic_dataset.py
innerlee: This can be done in a separate pr

PR #147, opened 2021-04-30
MMOCR COlab

quincylin1:
Colab for mmocr, including:

Training, testing and visualization for text recognition demo dataset
Testing and visualization for text detection with PANet
Testing and visualization for KIE


innerlee:
Linting fails

PR #150, opened 2021-05-03
Implement iter protocol for Loader

innerlee:
Otherwise, the loader is not compatible with for loop iteration.

### DIFF
--- a/mmocr/datasets/utils/loader.py
+++ b/mmocr/datasets/utils/loader.py
@@ -39,6 +39,17 @@ class Loader:
         """Retrieve anno info of one instance with dict format."""
         return self.parser.get_item(self.ori_data_infos, index)
 
+    def __iter__(self):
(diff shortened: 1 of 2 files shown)


mmocr/datasets/utils/loader.py
```diff
@@ -39,6 +39,17 @@ def __getitem__(self, index):
...
+    def __iter__(self):
+        self._n = 0
+        return self
+
+    def __next__(self):
+        if self._n < len(self):
+            data = self[self._n]
+            self._n += 1
+            return data
+        raise StopIteration
```
jeffreykuang, line 51: it will always lead to stopiteration exception as self._n keeps increasing?

innerlee: yeah this is intended

cuhk-hbsun (approved):

PR #159, opened 2021-05-08
fix bug of loss nan

cuhk-hbsun:
Fix ctc_loss NaN bug by setting `zero_infinity=True, flatten=False` for ctc_loss

### DIFF
--- a/mmocr/models/textrecog/losses/ctc_loss.py
+++ b/mmocr/models/textrecog/losses/ctc_loss.py
@@ -1,3 +1,5 @@
+import math
+
 import torch
 import torch.nn as nn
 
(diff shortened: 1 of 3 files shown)


mmocr/models/textrecog/losses/ctc_loss.py
```diff
@@ -54,6 +54,8 @@ def forward(self, outputs, targets_dict):
                 targets[idx, :valid_len] = tensor[:valid_len]
 
         target_lengths = targets_dict['target_lengths']
+        target_lengths = torch.clamp(
+            target_lengths, min=1, max=seq_len - 1).long()
```
jeffreykuang, line 58: How about max=seq_len instead of max=seq_len-1?

jeffreykuang:
Since zero_infinity=True, flatten=False is the key for avoiding nan loss. Why not set these value by default in the ctc loss initialization?  Is there any side effect if doing so?   If let users to set them in cofig.  I think it will result in difficulty of usage.

mmocr/models/textrecog/losses/ctc_loss.py
```diff
@@ -54,6 +59,16 @@ def forward(self, outputs, targets_dict):
                 targets[idx, :valid_len] = tensor[:valid_len]
 
         target_lengths = targets_dict['target_lengths']
+        target_lengths = torch.clamp(target_lengths, min=1, max=seq_len).long()
+
+        input_lengths = torch.full(
+            size=(bsz, ), fill_value=seq_len, dtype=torch.long)
+        if not self.flatten and valid_ratios is not None:
```
jeffreykuang, line 66: why self.flatten=False?  How about if it is true?

cuhk-hbsun: If flatten is True, `input_lengths` must be filled with the value `seq_len` to use cudnn.

mmocr/models/textrecog/losses/ctc_loss.py
```diff
@@ -36,12 +38,15 @@ def __init__(self,
         self.ctc_loss = nn.CTCLoss(
             blank=blank, reduction=reduction, zero_infinity=zero_infinity)
 
-    def forward(self, outputs, targets_dict):
+    def forward(self, outputs, targets_dict, img_metas=None):
```
jeffreykuang, line 41: When is the img_metas is None and when is it not None? How to control it? Can we set it in config?

jeffreykuang (approved):

PR #183, opened 2021-05-13
update metafile

cuhk-hbsun:


configs/textdet/textsnake/metafile.yml
```diff
@@ -1,7 +1,7 @@
 Collections:
 - Name: TextSnake
   Metadata:
-    Training Data: ICDARDataset
+    Training Data: SCUT-CTW1500
```
quincylin1, line 4: change to ICDAR SCUT-CTW1500 to be consistent with above

quincylin1 (approved):

PR #193, opened 2021-05-15
Create environment.yml

seekingdeep:


seekingdeep:
This file is to simplify the installation, and create a new python environment called `mmocr`.
`mmocr` will contain Python **3.7**/ Pytorch **1.5**/ torchvision **0.6**/ Cuda **10.1**/ mmcv-full **1.3.4**
1 command to install:
```
conda env create -f environment.yml
conda activate mmocr
```

innerlee:
Thanks for the contribution! However, creating an universally applicable env is very hard. E.g. in the line below, `cu101` maybe not applicable with user's environment. And `mmcv` is also continuously evolving. A good solution is to make the installation page (https://github.com/open-mmlab/mmocr/blob/main/docs/install.md) more friendly and helpful. What do you think?

image

PR #217, opened 2021-05-18
Test more on cpu

innerlee:
Signed-off-by: lizz <lizz@sensetime.com>


jeffreykuang (commented):
can we do the unit tests in both cpu and gpu modes

cuhk-hbsun (approved):

PR #244, opened 2021-05-27
add backend for resizeocr

cuhk-hbsun:


mmocr/datasets/pipelines/ocr_transforms.py
```diff
@@ -50,13 +54,15 @@ def __init__(self,
             assert isinstance(min_width, tuple)
             assert isinstance(max_width, tuple)
             assert len(height) == len(min_width) == len(max_width)
+        assert backend in ('cv2', 'pillow') or backend is None
```
innerlee, line 57: `None` could be part of the tuple

innerlee: Acutally we can remove this assertion. mmcv might support more backends

innerlee (approved):

PR #246, opened 2021-05-27
Fix key-value conflict of upsamle_layers between mmseg and mmocr.

RangiLyu:
To solve key-value conflict of upsample layers in unet between mmseg and mmocr, I add namespace when registering upsample layers.


cuhk-hbsun:
please run `pre-commit run --all-files` before `git commit`

PR #263, opened 2021-06-07
fix #253: speedup; fix assert

### LINKED ISSUE #253: Performance Issue about DBNet
plz ref to https://github.com/MhLiao/DB/pull/259.

cuhk-hbsun:


innerlee:
Any numbers on the speedup?

innerlee:
It seems that the changes contains some bug fix. Please add tests for them. The tests should fail before the change and pass after the change.

innerlee (approved):

PR #264, opened 2021-06-07
fix #113: textsnake decode error

### LINKED ISSUE #113: There is a chance for the contour area equals 0, which will raise error when training and testing.
I found the bug when testing the textsnake with my own dataset, the bug randomly occurred during training and testing. After serveral hours reading the code and paper, I managed to fix it by adding a check inside.
change this
https://github.com/open-mmlab/mmocr/blob/ce98d23ba2b5323348679dc19614d35c747c6a3c/mmocr/models/textdet/postprocess/wrapper.py#L390
to 
```python
if len(contours) > 0 and cv2.contourArea(contours[0]) > 0:
```
The contour for the boundary may be a single point sometime.

HolyCrap96:


innerlee:
Could you please add test case for these corner cases?

HolyCrap96:
Recall Precision H-mean
before modification: 
0.795 0.840 0.817
after modification:
0.795 0.841 0.817

cuhk-hbsun (approved):

PR #281, opened 2021-06-12
fix #279: save detect results

### LINKED ISSUE #279: saving the quadrilateral predicted .txt
hi,
how to save the quadrilateral predicted .txt

`sh tools/test_imgs.sh configs/textdet/psenet/psenet_r50_fpnf_600e_ctw1500.py ./epoch_20.pth ./input ./input/list.txt --save_quad true`

cuhk-hbsun:


tools/det_test_imgs.py
```diff
@@ -0,0 +1,114 @@
...
+        default='./results',
+        help='Dir to save '
+        'visualize images '
+        'and bbox')
+    args = parser.parse_args()
+
+    assert 0 < args.score_thr < 1
+
+    # build the model from a config file and a checkpoint file
+    device = 'cuda:' + str(torch.cuda.current_device())
```
innerlee, line 74: Can it support cpu?

cuhk-hbsun: Yes, except that `norm_cfg=dict(type='SyncBN')` in `backbone` of some detect models.
Add new commit setting device as args.

tools/recog_test_imgs.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
innerlee, line 1: I don't understand why this bash wrapper is needed.
Python scripts have better help tools. You can type `./tools/recog_test_imgs.py -h` to see which argument is needed.

cuhk-hbsun: unnecessary

innerlee (approved):

PR #285, opened 2021-06-15
fix #284: str only for separator

### LINKED ISSUE #284: LineStrParser: separator behaviour
I've a question regarding this snippet of code:
https://github.com/open-mmlab/mmocr/blob/01d8d63be945882fb2d9eaca5e1c1b39cb45f274/mmocr/datasets/utils/parser.py#L33-L36

Is there a particular reason to use these 4 lines of code instead of simply `line_str = line_str.split(self.separator)`?

I'm asking this because for my own use case I have:
- a TSV file with `filename` and `text` as keys for text recognition task
- some blank spaces in `filename` e.g. `my cropped image.png`
 
Hence, LineStrParser is configured as follows:
```python
parser=dict(
            type='LineStrParser',
            keys=['filename', 'text'],
            keys_idx=[0, 1],
            separator='\t'))
```
but with the 4-lines code snippet, the line parsing fails. Instead, with simply `line_str = line_str.split(self.separator)` everything works well.

cuhk-hbsun:


gaotongxiao (approved):

innerlee:
Note for future: for bug fixes, unittest for corner cases should be added. @gaotongxiao @cuhk-hbsun

PR #288, opened 2021-06-15
[Fix] Correct isort setup

AllentDan:
Hi, this is a PR for correcting isort setup.
Mainly changed:
modify mmdet to 'known_third_party' in setup.cfg and add tools to 'known_first_party' in setup.cfg


innerlee:
What's the reason for the previous "error"?

AllentDan:
@innerlee Not error, just code requirements....

Can be ignored acctually.

AllentDan:
@innerlee. I just want to add the following line:
```
from mmocr.datasets.pipelines.crop import crop_img # noqa: F401
```
to onnx2tensorrt.py. But found that mmdet should be the third party of this repo and tools should be added to the first party.

innerlee:
Okay, why `tools` folder matter? They are not packages

AllentDan:
The tool folder is imported in some of my PRs... But I forgot to add it to 'known_first_party' of this repo... If not specified, isort will move lines to 'known_third_party'.

Actually, move 'mmdet' to 'known_third_party' may cause some conflicts to somebody working on new PRs. So, I am not sure if it should be done in this PR. Maybe it should be done by the owner of this repo.

innerlee:
> The tools folder is imported in some of my PRs

This part is what I was concerning about. Can this "import" be avoided in some way?

> done by the owner of this repo.

We do not have such restrictions. Contributions of any form are always welcome.

AllentDan:
Hi, @innerlee. It can be avoided through replacing "import tools.deployment.something as sth" with "import .something as sth" or "import something as sth". What is your suggestion? Seems mmdet prefers the former.

innerlee:
tools are stand-alone scripts. If something are reusable, then they should live in the package. So my suggestion is to move deploy_helper inside `mmocr/`

AllentDan:
Hi, @innerlee. How about this. In this PR, I only do the following stuff:
1. move mmdet to 'known_third_party'.
2. add 'from mmocr.datasets.pipelines.crop import crop_img # noqa: F401' to onnx2tensorrt.py.

I will move deploy_helper.py to mmocr in the next PR which aims to evaluate exported onnx and tensorrt files. Related unit tests will be added then.

innerlee (approved):

innerlee:
fair enough.

Its better to remove `# noqa: F401` in later pr also.

AllentDan:
Hi, "# noqa: F401" was used because crop_img was not used. However, if not importing this func, a registry error will occur.

AllentDan:
Just like [here](https://github.com/open-mmlab/mmocr/blob/d57f279083a9fb63ad4a5665203cf9fbff9271f2/demo/batch_image_demo.py#L8)

innerlee:
> Just like here

Its easy to add some meaningless/unharmful ops for them like `assert crop_img is not None`.
The previous examples should be cleaned also.

Issue #282, opened 2021-06-12
Use init_cfg to unify the parameter initialization approaches


ZwwWayne:
**Describe the feature**
Support `init_cfg` to initialize the model parameters.

**Motivation**
The dependency of MMDet has been locked to v2.11.0 because of the in-compatibility of parameter initialization introduced in MMDetection v2.12.0. To finally migrate to the newest MMDetection and enjoys the newest features, it is time to migrate the model parameter initialization and use `init_cfg`.

**Related resources**
The upgrade of MMDetection: https://github.com/open-mmlab/mmdetection/pull/4750
The migration of MMDetection3D that will be merged at the end of this month: https://github.com/open-mmlab/mmdetection3d/pull/622
The PR in MMCV: https://github.com/open-mmlab/mmcv/pull/780

PR #299, opened 2021-06-18
fix #280: update coco reference

### LINKED ISSUE #280: ‘COCO’ object has no attribute 'cat_img_map'
**Checklist**

1. ‘COCO’ object has no attribute 'cat_img_map'

**Describe the bug**
When I run `./tools/dist_train.sh configs/textrecog/seg/seg_r31_1by16_fpnocr_toy_dataset.py work_dirs/seg 1`,it reports this error.
image

**Reproduction**

1. `./tools/dist_train.sh configs/textdet/maskrcnn/mask_rcnn_r50_fpn_160e_icdar2015.py work_dirs/mask 1`

**Bug fix**
Substitute `from pycocotools.coco import COCO` for `from mmdet.datasets.api_wrappers import COCO` in mmocr.datasets.icdar_dataset.py.

gaotongxiao:
#280

### DIFF
--- a/mmocr/datasets/icdar_dataset.py
+++ b/mmocr/datasets/icdar_dataset.py
@@ -1,7 +1,7 @@
 import numpy as np
+from mmdet.datasets.api_wrappers import COCO
 from mmdet.datasets.builder import DATASETS
 from mmdet.datasets.coco import CocoDataset
-from pycocotools.coco import COCO
 
 import mmocr.utils as utils
 from mmocr.core.evaluation.hmean import eval_hmean
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,7 +20,7 @@ line_length = 79
 multi_line_output = 0
 known_standard_library = setuptools
 known_first_party = mmocr
-known_third_party = PIL,Polygon,cv2,imgaug,lanms,lmdb,matplotlib,mmcv,mmdet,numpy,packaging,pyclipper,pycocotools,pytest,rapidfuzz,scipy,shapely,skimage,titlecase,torch,torchvision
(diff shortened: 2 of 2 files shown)


innerlee:
Can this be tested?

gaotongxiao:
It seems `mmdet=2.11.0` doesn't have the api wrapper. We need to finish #282 and update the dependency before applying this patch.

innerlee:
Any solution that work on mmdet=2.11.0?

gaotongxiao:
> Any solution that work on mmdet=2.11.0?
> 
> Or, is #280 caused by higher mmdet version?

It's just because we're using API names not provided by official COCO (https://github.com/open-mmlab/mmdetection/issues/2913#issuecomment-639893148). I guess there's no good solution for now.

innerlee:
So is this caused by the recent drop of mmpycocotools?

gaotongxiao:
> So is this caused by the recent drop of mmpycocotools?

Not a problem for `mmpycocotools`. Following the official installation instruction will install the official `pycocotools` instead of `mmpycocotools`.

innerlee:
Okay so there will be no way to make both mmpycocotools and pycocotools happy?

gaotongxiao:
> Okay so there will be no way to make both mmpycocotools and pycocotools happy?

Seems not. But we're working on #282 now so hopefully this issue can be fixed soon.

innerlee:
For maximum compatibility, I think you can inspect the existence of property `cat_img_map`, and if non-exist, then use the counterpart in pycocotools.

gaotongxiao:
> For maximum compatibility, I think you can inspect the existence of property `cat_img_map`, and if non-exist, then use the counterpart in pycocotools.

The problem comes from `CocoDataset`, which is a base class from MMDet. MMDet 2.11.0 assumes that the COCO object inside this class has `cat_img_map`, but it's not a standard property provided by pycocotools. We can copy the COCO wrapper in `MMDet 2.13.0` to MMOCR for maximum compatibility, but the safest way is still to upgrade the dependency and reuse their wrapper in case of any future change in their code.

Issue #303, opened 2021-06-20
Can't sucessfully run "Test a Single Image" demo


SWHL:
Thanks for your error report and we appreciate it a lot.

**Checklist**

1. I have searched related issues but cannot get the expected help.
2. The bug has not been fixed in the latest version.

**Describe the bug**
- I have already installed the relative environments by the [install.md](https://github.com/open-mmlab/mmocr/blob/main/docs/install.md)
- I run the **Test a single Image** demo according to the [getting_started.md](https://github.com/open-mmlab/mmocr/blob/main/docs/getting_started.md)
- I met the error: **ImportError: cannot import name 'stitch_boxes_into_lines' from 'mmocr.utils.img_util'**

**Reproduction**

1. What command or script did you run?

```shell
python demo/ocr_image_demo.py demo/demo_text_det.jpg demo/output.jpg
```

**Environment**
```log
sys.platform: linux
Python: 3.7.0 | packaged by conda-forge | (default, Nov 12 2018, 20:15:55) [GCC 7.3.0]
CUDA available: True
GPU 0,1,2,3: Tesla P100-PCIE-12GB
CUDA_HOME: /usr/local/cuda-10.0/
NVCC: Cuda compilation tools, release 10.0, V10.0.130
GCC: gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-39)
PyTorch: 1.6.0+cu101
PyTorch compiling details: PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
  - Intel(R) Math Kernel Library Version 2019.0.5 Product Build 20190808 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v1.5.0 (Git Hash e2ac1fac44c5078ca927cb9b90e1b3066a0b2ed0)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - NNPACK is enabled
  - CPU capability usage: AVX2
  - CUDA Runtime 10.1
  - NVCC architecture flags: -gencode;arch=compute_37,code=sm_37;-gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_60,code=sm_60;-gencode;arch=compute_70,code=sm_70;-gencode;arch=compute_75,code=sm_75
  - CuDNN 7.6.3
  - Magma 2.5.2
[...]

PR #304, opened 2021-06-20
[Fix] fix wrong import path

gaotongxiao:
#303

### DIFF
--- a/demo/ocr_image_demo.py
+++ b/demo/ocr_image_demo.py
@@ -6,7 +6,7 @@ from mmdet.apis import init_detector
 from mmocr.apis.inference import model_inference
 from mmocr.core.visualize import det_recog_show_result
 from mmocr.datasets.pipelines.crop import crop_img
-from mmocr.utils.img_util import stitch_boxes_into_lines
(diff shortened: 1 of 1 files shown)


innerlee:
Please add a test for running the script. See https://github.com/open-mmlab/mmcv/pull/1007 for how to do it

gaotongxiao:
> Please add a test for running the script. See [open-mmlab/mmcv#1007](https://github.com/open-mmlab/mmcv/pull/1007) for how to do it

Would it be too slow or even impossible to finish the tests since most demo scripts involve model inference?

innerlee:
Simply try it and fix it if run out of quota

gaotongxiao:
> Simply try it and fix it if run out of quota

This script is GPU-only, so I tested it locally and LGTM. We may consider adding tests for other demo scripts in the future.

innerlee:
> GPU-only

why

gaotongxiao:
> > GPU-only
> 
> why

An op (`SyncBatchNorm`) of its default detection model, PSENet, has to be on GPU

cuhk-hbsun (approved):

PR #310, opened 2021-06-21
support batch inference during testing

cuhk-hbsun:


docs/getting_started.md
```diff
@@ -325,4 +341,9 @@ You can check the content of the annotation file in `tests/data/toy_dataset/inst
   ```shell
   python tools/data_converter/ctw1500_converter.py ${src_root_path} -o ${out_path} --split-list training test
   ```
-```
+
+#### UniformConcatDataset
+
+To use the same pipeline for multiple datasets, as well as support extra parameters such as `samples_per_gpu` in `cfg.data.test` to activate `batch inference`, we design `UniformConcatDataset`. See [config](/configs/textrecog/sar/sar_r31_parallel_decoder_toy_dataset.py) for an example.
+
+If you want to use `batch inference` during testing, please update the config file refering to line108 - line117 in [config](/configs/textrecog/sar/sar_r31_parallel_decoder_toy_dataset.py)
```
innerlee, line 349: You can reference line no in the url

configs/textrecog/sar/sar_r31_parallel_decoder_toy_dataset.py
```diff
@@ -102,18 +102,17 @@
     test_mode=True)
 
 data = dict(
-    samples_per_gpu=16,
```
innerlee, line 105: if samples_per_gpu for train/val/test are same, then you can specify once at this line.

mmocr/apis/inference.py
```diff
@@ -6,23 +6,40 @@
 from mmdet.datasets.pipelines import Compose
 
 
-def disable_text_recog_aug_test(cfg):
+def disable_text_recog_aug_test(cfg, set_types=None):
     """Remove aug_test from test pipeline of text recognition.
     Args:
         cfg (mmcv.Config): Input config.
+        set_types (list[str]): Type of dataset source. Should be
```
innerlee, line 13: What does "type of dataset" mean?

configs/textrecog/sar/sar_r31_parallel_decoder_toy_dataset.py
cuhk-hbsun: 1. Currently global `data.samples_per_gpu` just for `train`, but not `val` and `test`. 
2. Since most of text detect models (use `MultiScaleFlipAug` in test_pipeline) and text recognition models (use `MultiRotateAugOCR` in test_pipeline) don't support batch inference.
3. The batch test will be activated only when setting `test_dataloader=dict(samples_per_gpu=8)`. It will disable `aug_test` for text recognition.

mmocr/apis/inference.py
cuhk-hbsun: `val` or `test`. For validation, it shpuld be `['val']`. For test, it should be `['test']`

configs/textrecog/sar/sar_r31_parallel_decoder_toy_dataset.py
innerlee: > will disable aug_test

why this is necessary?

innerlee: > Currently global data.samples_per_gpu just for train, but not val and test.

This story is too complex. An alternative way is to make samples_per_gpu a global option, and explicitly overwrite samples_per_gpu=1 in val/test. 

Special case random things feels wrong

cuhk-hbsun: `samples_per_gpu` be global setting

innerlee:
For safety, each modfied configs should be run once (for training).
Each type of models should run train.py/test.py/demo.py once.

We will add them to tests later. Before that, manual check is needed

innerlee (approved):

PR #319, opened 2021-06-23
Fix kie_test_imgs script for novisual

innerlee:
Fixed by loading img by hand if no img was there

I will add a test for this script later.


cuhk-hbsun:
need fix unittest

cuhk-hbsun (approved):

PR #323, opened 2021-06-24
update mmcv installation method

gaotongxiao:


gaotongxiao:
@zhouzaida

zhouzaida (approved):
LGTM

innerlee:
Reason for this change? Did we changed the mmcv lower version bound?

Historically, latest mmcv sometimes breaks packages, and the whole CI would stop work for several days. This is the reason we used a fixed version.

Adding a checkitem for the latest stable mmcv and the latest bleeding edge mmcv would be great. The current change does not gives improvement but introduces an old drawback.

innerlee (approved):

Issue #318, opened 2021-06-23
".txt" format ground truth file support for TOTAL-TEXT data 


mnansary:
# ** COCO-Format / Instances.json creation  from .txt format ground truth for Total-Text Format data **

**Motivation**
This feature will use ".txt" format ground truth data instead of  ".mat"  data.  
* While creating ```custom dataset```  in "total-text" format, it's inconvenient to use Matlab for creating the ground Truth FIles. 

**Related resources**
the task can be achieved through simple steps:
* The folder structure remains the same as it is in the documentation. But now under the **annotations folder**, instead of ".mat"  files there will be ".txt" files
* The format [here](https://github.com/mnansary/mmocr/blob/main/tools/data/textdet/totaltext_converter.py#L49) in line 49 and 57 ".mat" will change to ".txt"
*  Instead of using a mat file reader, like this in code 

```python
    data = scio.loadmat(gt_path)
    if split == 'training':
        data_polygt = data['polygt']
    elif split == 'test':
        data_polygt = data['polygt']
```

the  following part can be used:

```python

   if split == 'training':
        data_polygt=[]
        with open(gt_path,"r") as f:
            data_polygt=f.readlines()    
        #data_polygt = data['polygt']
    elif split == 'test':
        data_polygt=[]
        with open(gt_path,"r") as f:
            data_polygt=f.readlines()  
```
* The X, Y, and word data can be retrieved as follows

```python
    # change
    parts=lines.split(",")
    X=parts[0].split(":")[-1].strip().replace("[","").replace("]","").split()
    X=np.array([[int(x) for x in X]])
    
   # chnage
    Y=parts[1].split(":")[-1].strip().replace("[","").replace("]","").split()
    Y=np.array([[int(y) for y in Y]])
   

   point_num = len(X[0])
    
   # change
    word=[parts[-1].split(":")[-1].strip().replace("[","").replace("]","").replace("u'","").replace("'","")]
```
**Additional context**
I can create a separate file for handling  ".txt" format ground truth

cuhk-hbsun:
Thanks a lot. You can contribute by a pull request to modify the related scripts to support both `mat` and `txt` formats, with additional args like:
`parser.add_argument('label_format', help='Format of label format, can be either "txt" or "mat"')`

gaotongxiao:
[...]

PR #324, opened 2021-06-24
Issue 314

quincylin1:


mnansary:
Wow I have been working with same problem [here](https://github.com/open-mmlab/mmocr/issues/318) .. Should I proceed to close my issue? code : https://gist.github.com/mnansary/2c052d9ef0301295757526e1d8b3e463

Issue #271, opened 2021-06-09
When I want to train a fcenet, I met a problem


ShiMinghao0208:
![QQ图片20210609200559](https://user-images.githubusercontent.com/53176521/121351451-5f6cb280-c95e-11eb-982b-b7c65345bb39.png)
It did run, but several epoch,there note that data not on the gpu.
how to solve it?

cuhk-hbsun:
@ShiMinghao0208 Can you try again and see if the same problem arises?

ShiMinghao0208:
> @ShiMinghao0208 Can you try again and see if the same problem arises?

Yes, I ran the code several times and this problem came up every time

Zyq-scut:
@ShiMinghao0208 I create an environment following the guidance and train FCENet. It work well and doesn't raise the error you mentioned. Can you check if the same problem arises when train other models? If so, i think it is caused by environment.

innerlee:
Maybe it is caused by some bad data? Try to use binary search to locate

whynot08:
@ShiMinghao0208 Have you solved the problem yet? I met the same situation

ShiMinghao0208:
> @ShiMinghao0208 Have you solved the problem yet? I met the same situation

The guidance have an error, the mmcv is cpu version.When I install the gpu-mmcv,the error disappeared.

whynot08:
> > @ShiMinghao0208 Have you solved the problem yet? I met the same situation
> 
> The guidance have an error, the mmcv is cpu version.When I install the gpu-mmcv,the error disappeared.

I will have a try, thank you very much.

innerlee:
@ShiMinghao0208 could you please share the location of improper guidance? We will fix it asap

ShiMinghao0208:
> @ShiMinghao0208 could you please share the location of improper guidance? We will fix it asap

I dont know the location.I asked others for help, and he said that was the reason

innerlee:
Thanks for the feedback!

mattlee2:
I am having this exact same problem. I installed the MMCV that matches my Torch/CUDA.  What's even more peculiar is that I only get this error when training on certain datasets.  For others, it appears to begin training properly.

innerlee:
@mattlee2 It would be great if the error could be reproduced.

it might be due to some imgs in the dataset triggered some corner cases in the code. so

> Try to use binary search to locate the bad data

FFoCC:
[...]

PR #331, opened 2021-06-29
[Docs] update dependency version req, dockerfile and change logs for 0.2.1

gaotongxiao:
Docs & dockerfile update for 0.2.1


innerlee:
Is there a related issue in pytorch's repo?

gaotongxiao:
https://github.com/pytorch/pytorch/issues/33870

innerlee:
Reading the pytorch thread, it seems that it is caused by op on mixing devices. At #271, there is a cpu and a cuda

image

It seems that this could be avoided if we write code in fcenet more careful, correct?

gaotongxiao:
Yes. It's the quickest way to "fix" this issue though.

innerlee:
I'm not a fan of upgrading dependencies. It means that previously working code will refuse to run after upgrading. This will hurt old users.

Fixing the error from the code side would be great.

gaotongxiao:
I'll create a PR for init_cfg support soon which would also upgrade dependencies. By that time the results of the retrianing process can better prove that this upgrade is safe. We can hold the PR till then.

gaotongxiao:
To be merged together with the release of 0.2.1

gaotongxiao:
Date & version are subject to change. Will review 2 community PRs then and decide if they can be merged into this release.

cuhk-hbsun (approved):

PR #343, opened 2021-07-01
use unifor_concat_datase instead of concat_dataset

cuhk-hbsun:
Use `uniform_concat_dataset` such that `cfg.data.val` and `cfg.data.test` have attribute `pipeline`.


innerlee (approved):

innerlee:
For this behavior change, pls add a test

image

PR #348, opened 2021-07-05
Added support for ConcatDataset with two workflows

antoniolanza1996:
This PR added support for `ConcatDataset` with two workflows. In particular, if one wants to use `workflow = [('train', 1), ('val', 1)]` but in the configuration file there is a `ConcatDataset`, the actual `tools/train.py` is not able to handle it. In this case, as a hotfix, I propose to set the pipeline for each dataset in `datasets`. 

Note: I have seen the new `UniformConcatDataset`'s implementation and in the future I will use this new MMOCR component, but for the people that still continue to use `ConcatDataset`, I suppose this PR could help.

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -186,7 +186,16 @@ def main():
     datasets = [build_dataset(cfg.data.train)]
     if len(cfg.workflow) == 2:
         val_dataset = copy.deepcopy(cfg.data.val)
-        val_dataset.pipeline = cfg.data.train.pipeline
+        if cfg.data.train['type'] == 'ConcatDataset':
+            train_pipeline = cfg.data.train['datasets'][0].pipeline
+        else:
+            train_pipeline = cfg.data.train.pipeline
+
+        if val_dataset['type'] == 'ConcatDataset':
+            for dataset in val_dataset['datasets']:
+                dataset.pipeline = train_pipeline
+        else:
+            val_dataset.pipeline = train_pipeline
         datasets.append(build_dataset(val_dataset))
     if cfg.checkpoint_config is not None:
         # save mmdet version, config file content and class names in


cuhk-hbsun:
@antoniolanza1996 Thanks for your contribution. We have used `UniformConcatDataset` instead of `ConcatDataset` in [343](https://github.com/open-mmlab/mmocr/pull/343), so that both `cfg.data.train` and `cfg.data.val` have attribute `pipeline`.  
Just delete [line189](https://github.com/open-mmlab/mmocr/pull/348/files#diff-7fc3906bb14652119559420a7b84aac995bfa36f9a7b0f00e215b9f0b6301d9bL189) will be fine.

antoniolanza1996:
Hi @cuhk-hbsun, `val_dataset.pipeline = cfg.data.train.pipeline` is setting the training pipeline into validation dataset. I think if we delete it, we are not able to compute validation loss with `workflow = [('train', 1), ('val', 1)]`.

cuhk-hbsun:
> Hi @cuhk-hbsun, `val_dataset.pipeline = cfg.data.train.pipeline` is setting the training pipeline into validation dataset. I think if we delete it, we are not able to compute validation loss with `workflow = [('train', 1), ('val', 1)]`.

You are right. You can test the latest code which using `UniformConcatDataset`. If no problem, you can close this pr.

antoniolanza1996:
> You can test the latest code which using UniformConcatDataset

Sure, I'll give it a shot ASAP.

> If no problem, you can close this pr.

In other [tools](https://github.com/open-mmlab/mmocr/tree/20f65fb521c14095d2264967482dc6b76e4a54c8/tools) and [demo](https://github.com/open-mmlab/mmocr/tree/20f65fb521c14095d2264967482dc6b76e4a54c8/demo) scripts (e.g. [here](https://github.com/open-mmlab/mmocr/blob/20f65fb521c14095d2264967482dc6b76e4a54c8/tools/det_test_imgs.py#L78), [here](https://github.com/open-mmlab/mmocr/blob/20f65fb521c14095d2264967482dc6b76e4a54c8/tools/recog_test_imgs.py#L63), [here](https://github.com/open-mmlab/mmocr/blob/20f65fb521c14095d2264967482dc6b76e4a54c8/demo/batch_image_demo.py#L32)) there is still `ConcatDataset` reference, for this reason I propose this PR. Indeed, I think there are a lot of people in the MMOCR community that have their own custom configurations with `ConcatDataset` inside and if we definitively remove  `ConcatDataset` support this will break their code.

Alternatively you can think to leave these `ConcatDataset` references in `demo` and `tools` scripts (and then also this PR) for some months and also put some deprecation warnings (e.g. "`ConcatDataset` is no longer used in MMOCR. Use `UniformConcatDataset` instead") to make the deprecation process smoother, what do you think?

cuhk-hbsun:
I agree with you. Thanks a lot.

cuhk-hbsun (approved):

PR #351, opened 2021-07-06
Add synthtext converter and update docs

gaotongxiao:
The script is bulit based on the work of @yuexy

### DIFF
--- a/docs/datasets.md
+++ b/docs/datasets.md
@@ -267,21 +267,27 @@ The structure of the text detection dataset directory is organized as follows.
 
 - For `SynthText`:
   - Step1: Download `SynthText.zip` from [homepage](https://www.robots.ox.ac.uk/~vgg/data/scenetext/)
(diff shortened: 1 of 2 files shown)


docs/datasets.md
```diff
@@ -267,22 +267,28 @@ The structure of the text detection dataset directory is organized as follows.
...
-  - Step4:
+  - Step2:
 
   ```bash
+  mkdir SynthText && cd SynthText
+  mv /path/to/SynthText.zip SynthText
   unzip SynthText.zip
-
-  cd SynthText
+  mv SynthText synthtext
```
innerlee, line 276: why not mkdir synthtext directly?

gaotongxiao: To be compatible with the creepy shuffle_labels.txt :P

gaotongxiao: to be compatible with the dataset structure presented above

cuhk-hbsun (approved):

PR #361, opened 2021-07-12
fix syntext_converter

gaotongxiao:
Added integer casting to avoid potential breaking


cuhk-hbsun (approved):

innerlee:
Gentle reminder for future bug fixes: pls add a test case to make sure that the bug will no longer return

Issue #335, opened 2021-06-29
Inference speed is abnormally slow


samayala22:
(All the files I will be referencing are in this google [drive](https://drive.google.com/drive/folders/1AsFeGLUnk4XCpYoetHJLMqHrxZMCReNJ?usp=sharing))

I wrote a user friendly API interface so that people could try and put into production the algorithms in mmocr much more easily (ocr_main.py). Here's an example:

```python
from ocr_main import MMOCR

ocr = MMOCR(textdet= 'DB_r18', textrecog='CRNN')
results = ocr.readtext('.\\demo\\demo_text_det.jpg',out_img='.\\demo\\testing_out.jpg',details=False)
print(results)
```

Unfortunately, when testing (test.py), the inference speed was very slow. About 5 times slower compared to the PaddlePaddleOCR DB+CRNN implementation. When using DB+NRTR in mmocr,  it took 10 min to recognise the demo_text_det.jpg 50 times !

- mmocr_DB_CRNN.prof : profiling results of my test script
- ucr_DB_CRNN.prof : profiling results of the [UCR](https://github.com/DocYard-ai/UCR) implementation which uses PaddlePaddle's scripts.

As you can see when comparing the profiling results, mmocr is making too many calls to some pytorch built-in methods.

So I'd like to know if something is wrong with my interface, or is it because of the mmocr implementation or maybe something else ?

Once the issue is solved, I will create a PR to submit my ocr_main.py file.

_Note: Put the ocr_main.py and test.py files in the mmocr/mmocr directory (same dir where there is apis, core, datasets,...) and call the test.py script from the root mmocr directory (because mmdet is sensitive to pwd)._

Testing environment:
- Windows 10 (yes I managed to compile lanms)
- i7 9750h and RTX 2060
- Python 3.9
- Pytorch 1.8
- Cuda 11.1

gaotongxiao:
Sorry for the late reply. I have tested your script on my machine (ubuntu 16.04, gtx 1060, i7 8700), and it took ~15s on DB+CRNN, 3.5 min on DB+NRTR. There is performance loss on Windows since MMCV does not officially support windows yet.
As for UCR, I wasn't able to download their pretrained models due to the broken links. 
We have already noticed the inference time problem. Currently, we already have a plan to remove some redundant pipelines; also, we started to provide some deployment options aiming at reducing the inference speed. We will also analyze your .prof and try to reduce unnecessary calls.

samayala22:
[...]

PR #366, opened 2021-07-13
Add User friendly API

samayala22:
Adds a file that can be called from the **command line** or by **calling the module from an external script** and performs end to end ocr much more easily than using the demo file. In my opinion, this would make mmocr much more accessible to beginners and people who are interested in quickly trying out the implemented algorithms.

Example:
```python
from ocr import MMOCR

ocr = MMOCR(textdet='MaskRCNN_ICDAR17', textrecog='SEG')
results = ocr.readtext('.\\demo\\demo_text_recog.jpg', out_img='.\\demo\\out.jpg', details=False)
```

or

`python ocr.py demo\demo_text_recog.jpg --textdet PS_ICDAR15 --textrecog CRNN --imshow --details`

All the arguments are pretty much self-explanatory. 
The detection and recognition models are listed at the beginning of the file.

The API has been designed to be easily modified and adding arguments is as easy as adding the new arg in the `parse_args` method, adding the new arg in the arguments of the `readtext` method inside the `MMOCR` class. You can now access the value of the argument with `args.<new_arg>` inside the `readtext` method.

I have already mentioned the performance issues in [#335](https://github.com/open-mmlab/mmocr/issues/335) (but nobody addressed it) and I'm pretty sure this has something to do with the implementation of the models and not the API.

_I'm not a programmer and I didn't fully understand all the contributing guidelines steps, my apologies in advance if this PR is not formatted correctly._


gaotongxiao (commented):
Thanks for the nice work! That would be really helpful for users. Just a few comments:
1. Is it necessary to rename the config files? If not, sticking to original names is safer as it will not break the code;
2. The script could be moved to `tools/` 
3. It would be great if you also add a short intro section to getting_started.md (https://mmocr.readthedocs.io/en/latest/getting_started.html#useful-tools). If that's too much for you, we can also do it later.
4. There might be some minor problems that cannot pass our linting test (e.g. the file doesn't end with a new line) but can be fixed by pre-commit hook. See (https://github.com/open-mmlab/mmocr/blob/main/.github/CONTRIBUTING.md#python and https://github.com/open-mmlab/mmocr/blob/main/.github/CONTRIBUTING.md#step-3-commit-your-changes for details). In short, try
```
pre-commit install
pre-commit run --all-files
```

samayala22:
Thanks for the quick response !

1. Not it's not necessary, I did it to improve consistency between the repo naming structure and the one in the openmmlab server: `https://download.openmmlab.com/mmocr/textrecog/robustscanner/robustscanner_r31_academic-5f05874f.pth`. But I understand the concerns and will stick with the original name of the config directory.
2. Alright, I tried it earlier but I kept getting `ModuleNotFoundError: No module named 'mmocr'` and after trying a lot of possible solutions unsuccessfully I just placed it in the root directory. I will try again tomorrow.
3. Got it, will do it tomorrow.
4. Noted, I'll do it too.

tools/ocr.py
```diff
@@ -1,3 +1,7 @@
+import sys
+
+sys.path.insert(1, '../mmocr/')
+
```
gaotongxiao, line 4: Removing this should be fine

tools/ocr.py
```diff
@@ -0,0 +1,275 @@
...
+            self.device = device
+
+        if self.td not in textdet_models:
+            raise ValueError(
+                self.td, 'is not a supported text detection algorthm')
+        elif self.tr not in textrecog_models:
+            raise ValueError(
+                self.tr, 'is not a supported text recognition algorithm')
+
+        dir_path = os.path.dirname(os.path.realpath(__file__))
```
gaotongxiao, line 214: Use `dir_path = os.getcwd()`. We always assume user runs the script like
```
python tools/ocr.py demo/demo_text_det.jpg --imshow
```

gaotongxiao (commented):
I found it work well without the extra path insertion lines. Has your mmocr been correctly installed, or you are just running it as a standalone repo?

samayala22:
> I found it work well without the extra path insertion lines. Has your mmocr been correctly installed, or you are just running it as a standalone repo?

mmocr does not install on windows so I just cloned the repo in my documents directory. I guess it should work fine on linux then.

tools/ocr.py
samayala22: Wouldn't it be more flexible if we get the absolute path of the config files based on the location of the ocr.py file ? So that the user can run the script from whatever cwd directory he's in. Or will that break other things ? 
Also, why not put the file in the mmocr/utils directory so that the user can call it from any python script with :

```python
from mmocr.utils.ocr import MMOCR
```

Additionally, I believe the config folder should maybe be included in the installation setup (currently excluded). Maybe you can shine some light on the structural choices that have been made ? Because I don't see another way to make mmocr run easily out-of-the-box without extra steps.

gaotongxiao: > Wouldn't it be more flexible if we get the absolute path of the config files based on the location of the ocr.py file ? So that the user can run the script from whatever cwd directory he's in. Or will that break other things?
> ...
> Additionally, I believe the config folder should maybe be included in the installation setup (currently excluded). Maybe you can shine some light on the structural choices that have been made? Because I don't see another way to make mmocr run easily out-of-the-box without extra steps.

Including it into the package is not suggested because config files are more flexible than code and user should be able to modify them. However, it is possible to get the relative path of config files given your script file location. (btw, it might also be cool to support customized config path in case users want to test their own models)

> Also, why not put the file in the mmocr/utils directory so that the user can call it from any python script with :
> 
> ```python
> from mmocr.utils.ocr import MMOCR
> ```
Good idea. Also don't forget to modify `mmocr/utils/__init__.py` so that `from mmocr.utils import MMOCR` also works.

samayala22: I moved the file to `mmocr/utils`, modified the `mmocr/utils/__init__.py` and added arguments for custom detection and recognition configs. If there's anything else you think I should add as an argument let me know. Otherwise I will start writing something for the docs.

samayala22:
I moved the file to `mmocr/utils`, modified the `mmocr/utils/__init__.py` and added arguments for custom detection and recognition configs. If there's anything else you think I should add as an argument let me know. I was thinking of adding support for
- Choosing folder of images to run inference on
- Choose between det + recog, det only or recog only

gaotongxiao:
The first and second options sound really great! For the third one, I think supporting .json file is enough cuz we don't really have to support every possible format. Yet it would be a nice bonus.

In fact, the script looks good enough to be merged. But your codebase is a bit outdated, so it would be great to locally resolve the conflict with the latest main branch first.

mmocr/utils/__init__.py
```diff
@@ -9,6 +9,7 @@
 from .img_util import drop_orientation, is_not_png
 from .lmdb_util import lmdb_converter
 from .logger import get_root_logger
+from .ocr import MMOCR
 from .string_util import StringStrip
```
gaotongxiao, line 13: Try switching their order to pass the test (could avoid circular import)

samayala22: I tried to change it but when I try to commit the change `isort` automatically changes it
```
flake8...................................................................Passed
seed isort known_third_party.............................................Passed
isort....................................................................Failed
- hook id: isort
- files were modified by this hook

Fixing mmocr\utils\__init__.py
```

gaotongxiao (approved):
LGTM though circular import still doesn't get fixed. We may change the directory structure in the future.

gaotongxiao:
BTW would you plan to write something about your script in the docs?

samayala22:
Yep, I will write something. Where do you think is best ? 

My script is basically an improved version of the demo files because it combines them in a single file and adds some functionalities. So, perhaps I should change the demo tab in the docs ? And maybe delete the superseded demo files ?

gaotongxiao:
Great! You can just change the content in this file https://github.com/open-mmlab/mmocr/blob/main/demo/docs/ocr_demo.md. Also, `ocr_image_demo.py` is not needed and can be deleted now

OpenMMLab-Coodinator:
Hi @samayala22  ！First of all, we want to express our gratitude for your significant PR in this project. Your contribution is highly appreciated, and we are grateful for your efforts in helping improve this open-source project during your personal time. We believe that many developers will benefit from your PR.

We would also like to invite you to join our Special Interest Group (SIG) private channel on Discord, where you can share your experiences, ideas, and build connections with like-minded peers. To join the SIG channel, simply message moderator— OpenMMLab on Discord or briefly share your open-source contributions in the #introductions channel and we will assist you. Look forward to seeing you there! Join us ：https://discord.gg/UjgXkPWNqA

If you have WeChat account，welcome to join our community on WeChat. You can add our assistant ：openmmlabwx. Please add "mmsig + Github ID" as a remark when adding friends：）
Thank you again for your contribution❤

PR #368, opened 2021-07-15
seg .pt格式转换

Whatsetsthisend:


cuhk-hbsun:
@Whatsetsthisend Thanks for your contribution. There are some points you should pay attention to:
1. please sign the cla license first;
2. It will be better to create a feature branch in your forked repo to develop features instead of the `main` branch.
3. Please make pr to `main` branch instead of `zz/script`
4. Last but not least, why you want to remove the decoding part of `mmocr/models/textrecog/convertors/seg.py` and make change to `mmocr/models/textrecog/recognizer/base.py`? Is there any bug or something wrong?

Whatsetsthisend:
Because I need to use the torch.jit.trace function, the principle of the function is to use variables to cycle the whole prediction process, but it does not support forward functions with multiple parameters. I can only modify seg.py and base.py to delete img_ Measures parameter, and the introduction of Python constant in torch.jit.trace will cause the weight to freeze at the constant and cannot be changed, so the final translation part of the result is deleted, which needs to be implemented manually in C + +.

> @Whatsetsthisend感谢您的贡献。您应该注意以下几点：
> 
> 1. 请先签署 cla 许可证；
> 2. 最好在你的分叉仓库中创建一个功能分支来开发功能而不是`main`分支。
> 3. 请让 pr`main`分支而不是`zz/script`
> 4. 最后但并非最不重要的一点是，为什么要删除 的解码部分`mmocr/models/textrecog/convertors/seg.py`并对其进行更改`mmocr/models/textrecog/recognizer/base.py`？是否有任何错误或错误？

Because I need to use the torch.jit.trace function, the principle of the function is to use variables to cycle the whole prediction process, but it does not support forward functions with multiple parameters. I can only modify seg.py and base.py to delete img_ Measures parameter, and the introduction of Python constant in torch.jit.trace will cause the weight to freeze at the constant and cannot be changed, so the final translation part of the result is deleted, which needs to be implemented manually in C + +.

Whatsetsthisend:
The result of this transformation is the same as that of Python in C + + after my test

cuhk-hbsun:
@Whatsetsthisend I got it. You can create a new class in a new file that inherits the default class instead of modifying the original class directly.

configs/textrecog/seg/convert.py
```diff
@@ -0,0 +1,48 @@
+import torch
```
innerlee, line 1: The file is a tool, so it could be put at `tools/deployment/pytorch2torchscript.py`

gaotongxiao:
It seems this PR has had no update for several months. It is also a bit deviated from our development plan and we may not have enough time to perfect the details. Therefore, we decided to close this dangling PR for now. Thanks for your contribution anyway!

PR #371, opened 2021-07-17
User friendly API v2 + Docs !

samayala22:
This is a follow up to #366 

I spent way too much time on this but it was worth it !

Changelog:

- Added support for folder input, np array, list/tuple of np arrays or strings (image paths)
- Added batch_mode support for end2end ocr and det/recog inference.
- Added batch size argument
- Changed the result export method, you can now choose between different file formats (json, yml, pickle) and specify the directory where the files will be saved to.
- Streamlined the whole pipeline with some significant refactoring to improve code robustness and readability.
- Bug fixes and performance improvements.

... and added the new demo DOCS! This is my first time writing in Markdown so any suggestion for improvement is appreciated.

Removed the old .md files. Webcam demo is useless imo because inference is too slow anyways and the user can write his own pipeline in 5 lines of code.

 I kept the demo py files until a chinese version of this new doc is written.

I noticed that not all algorithms support batch mode correctly. They either return a resize error, only 1 result or another error. I think this issue should be addressed.


gaotongxiao (commented):
Thanks for the amazing work! 

For completeness of the documentation, I would suggest rewriting old demo commands  (which you have deleted) in your docs so that users can know how det and recog models work clearly. Old example figures can also be reused (or updated if you have a different UI). As a user, I would prefer docs with more figures.

Also, it might be better to copy two examples in your current docs to getting_started.md rather than asking users (much likely as beginners) to read the full docs. I believe your provided examples are straightfoward for a beginner to get hands on experience.

samayala22:
Thanks !

> I would suggest rewriting old demo commands (which you have deleted) in your docs so that users can know how det and recog models work clearly

By that, do you mean putting these old commands ?
```
python demo/image_demo.py ${TEST_IMG} ${CONFIG_FILE} ${CHECKPOINT_FILE} ${SAVE_PATH} [--imshow] [--device ${GPU_ID}]
```

I don't see how it explains how the models work more clearly ? Additionally, those old demo file might get deleted some day. I can do a flowchart (and insert it as an img in the docs) on how the script works if that's something you think could be useful.

> Old example figures can also be reused (or updated if you have a different UI). As a user, I would prefer docs with more figures.

My examples are different than in the old docs and I don't see where I could put useful visualization figures in the new docs. That's because I don't have separate sections for det+recog or det/recog only, as everything is done with the same script. I can however, put the expected output from the first example if you want.

gaotongxiao:
> I don't see how it explains how the models work more clearly ? Additionally, those old demo file might get deleted some day. I can do a flowchart (and insert it as an img in the docs) on how the script works if that's something you think could be useful.

Oh no, we don't need a flowchart. I mean some beginners might not know the difference between text detection and recognition, and the old demo files might explain that clearly. My suggestion is to add two more examples to the beginning of the docs, for running det only and recog only, preferably with example figures. It might better illustrate the idea. 

If the old figures can't be reused, you can also delete them.

samayala22:
@gaotongxiao Okay, it's done ! Hopefully that's what you had in mind.

gaotongxiao (approved):
Good work!

gaotongxiao:
We need some time to support readthedocs, don't merge now

gaotongxiao:
Just pushed some changes to fix some docs issues.
1. Unified the file structure by moving the docs from 'demo/docs' to `docs/` 
2. Renamed the title to Demo
3. Added commands for readthedocs that convert unsupported symbols to texts
4. Minor docs content fixes in support of readthedocs

PR #382, opened 2021-07-21
[Fix] upgrade version requirement of mmdet to 2.14.0 to avoid known bugs

gaotongxiao:
In mmdet 2.13.0, the `__init__()` function of `SingleStageDetector` always assign a value to the `pretrained` variable of the backbone, which could cause an error if the backbone does not accept `pretrained` in its initializer. (https://github.com/open-mmlab/mmdetection/pull/5363). It was fixed in `2.14.0`.

We upgrade the dependency requirement to avoid potential errors.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -15,7 +15,7 @@ RUN apt-get update && apt-get install -y git ninja-build libglib2.0-0 libsm6 lib
 RUN conda clean --all
 RUN pip install mmcv-full -f https://download.openmmlab.com/mmcv/dist/cu101/torch1.6.0/index.html
 
-RUN pip install mmdet==2.13.0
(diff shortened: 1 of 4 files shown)


innerlee (approved):

innerlee:
The recent dependency dance is a bit aggressive. It will break users' code. If you got time, it would be great to come up with a migration guide that points out common upgrade details. We will put it on doc.

innerlee:
Actually every version of any software has bugs. I would not suggest upgrade dependency versions so frequently next time.

gaotongxiao:
Sure, but the situation is special this time as we got BC-breaking changes in MMCV and MMDetection. MMDetection 2.13.0 was not that stable which can still break our code.

I agree the migration guide is missed and I will write one soon.

PR #400, opened 2021-07-29
[Feature] Improve ocr.py

### LINKED ISSUE #397: Allow a way to specify configs dir globally
User should be able to either provide the various functions/classes with a `config_dir`, or use an `export MMOCR_CONFIG_DIR`, or something like `mmocr.register_config_dir(path: str)`.

I would then grab all the configs and put them in `home/.mmocr_configs` and not have to worry about putting it in my projects.

Another use case might be that the project already has a `configs` folder used for another purpose and the user wishes to have mmocr configs in `mmocr_configs` instead.

If a user wishes to have a separate config folder for their own custom configs, in those cases they can explicitly specify it on a use-by-use basis.

**Motivation**
Working with the mmocr (or mmdet for that matter) in multiple projects can be a pain as one needs to copy the config into their project folder over and over. Moreover, it's not clear that it makes sense to use the name `configs` for the mmocr configs.

**Additional context**

One example is with `mmocr.utils.ocr.MMOCR`. There you can provide a `det_config` and a `recog_config` but you always have to change it depending on your `det` and `recog`. Could make it that a `config_dir` is provided without having to specify the exact file. But I think the env variable or global registration would be better.

gaotongxiao:
1. Add box stitching back to ocr.py
2. Add config_dir which allows users to specify the default config path
3. Warn users when overriding parameters are set
4. Allow users to use customized checkpoint files

Fix #397

### DIFF
--- a/mmocr/utils/ocr.py
+++ b/mmocr/utils/ocr.py
@@ -1,4 +1,5 @@
 import os
+import warnings
 from argparse import ArgumentParser, Namespace
 from pathlib import Path
 
@@ -9,6 +10,7 @@ from mmdet.apis import init_detector
 from mmocr.apis.inference import model_inference
 from mmocr.core.visualize import det_recog_show_result
 from mmocr.datasets.pipelines.crop import crop_img
+from mmocr.utils.box_util import stitch_boxes_into_lines
 
 textdet_models = {
     'DB_r18': {
@@ -167,10 +169,10 @@ def det_and_recog_inference(args, det_model, recog_model):
             box_res['box_score'] = float(bbox[-1])
             box = bbox[:8]
             if len(bbox) > 9:
-                min_x = min(bbox[0::2])
-                min_y = min(bbox[1::2])
-                max_x = max(bbox[0::2])
-                max_y = max(bbox[1::2])
+                min_x = min(bbox[0:-1:2])
+                min_y = min(bbox[1:-1:2])
+                max_x = max(bbox[0:-1:2])
+                max_y = max(bbox[1:-1:2])
                 box = [min_x, min_y, max_x, min_y, max_x, max_y, min_x, max_y]
             box_img = crop_img(arr, box)
             if args.batch_mode:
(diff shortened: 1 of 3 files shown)


mmocr/utils/ocr.py
```diff
@@ -166,10 +167,10 @@ def det_and_recog_inference(args, det_model, recog_model):
             box_res['box_score'] = float(bbox[-1])
             box = bbox[:8]
             if len(bbox) > 9:
-                min_x = min(bbox[0:-1:2])
-                min_y = min(bbox[1:-1:2])
-                max_x = max(bbox[0:-1:2])
-                max_y = max(bbox[1:-1:2])
```
samayala22: Are you sure that this is an indexing bug ? By removing the `-1` you are including the box score in one of the new variables which should only include the bounding shape coordinates. 

I might add that this whole part should be reworked because it does not make good use of the arbitrary shaped bounding regions returned by some detection models. Something should be created to pre-process those coordinates with a combination of automated warping and perspective projections to return the canonical form of the text. This would make recognition much more reliable on curved and inclined text.

mmocr/utils/ocr.py
```diff
@@ -167,10 +169,10 @@ def det_and_recog_inference(args, det_model, recog_model):
...
             box = bbox[:8]
             if len(bbox) > 9:
-                min_x = min(bbox[0::2])
-                min_y = min(bbox[1::2])
-                max_x = max(bbox[0::2])
-                max_y = max(bbox[1::2])
+                min_x = min(bbox[0:-1:2])
+                min_y = min(bbox[1:-1:2])
+                max_x = max(bbox[0:-1:2])
+                max_y = max(bbox[1:-1:2])
```
samayala22: I might add that this whole part should be reworked one day because it does not make good use of the arbitrary shaped bounding regions returned by some detection models. Something should be created to pre-process those coordinates with a combination of automated warping and perspective projections to return the canonical form of the text. This would make recognition much more reliable on curved and inclined text.

mmocr/utils/ocr.py
```diff
@@ -9,6 +10,7 @@
 from mmocr.apis.inference import model_inference
 from mmocr.core.visualize import det_recog_show_result
 from mmocr.datasets.pipelines.crop import crop_img
+from mmocr.utils.box_util import stitch_boxes_into_lines
```
samayala22, line 13: I originally didn't include that function because I personally find it badly designed and the result is clunky. I might submit a PR for a better solution when I'll have some spare time.

gaotongxiao: Agree, it's an experimental utility and definitely needs a better design and implementation. We are open to better solutions at any time.

cuhk-hbsun (approved):

PR #406, opened 2021-08-02
[Docs] Fix API reference & version, support exporting pdf & epub

gaotongxiao:
Since the introduction of dependency version checking in `__init__`, the old readthedocs config is no longer valid as it mocks MMCV and MMDetection and hides the version information in these packages. The assertion error blocks `autodoc` from importing modules and generate docs for APIs.

`autodoc_mock_imports` also unnecessarily mocked too many modules and this PR removes them.

Also modified code so that the version of the docs can be in sync with MMOCR.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -23,7 +23,7 @@ copyright = '2020-2030, OpenMMLab'
 author = 'OpenMMLab'
 
 # The full version, including alpha/beta/rc tags
-release = '0.1.0'
+release = '0.2.1'
 
 # -- General configuration ---------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


docs/conf.py
```diff
@@ -23,7 +23,7 @@
 author = 'OpenMMLab'
 
 # The full version, including alpha/beta/rc tags
-release = '0.1.0'
+release = '0.2.1'
```
zhouzaida, line 26: refer to https://github.com/open-mmlab/mmcv/blob/285a052ab7fdb7f37e9b1450cb3ab63748eab601/docs/conf.py#L22

cuhk-hbsun (approved):

PR #407, opened 2021-08-03
support batch inference for crnn and segocr

cuhk-hbsun:
support batch inference for all the resize modes in text recognition algorithms, including:
1. Resize to a fix shape;
2. Keep aspect ratio resize and pad to a fix shape; 
3. keep aspect ratio resize, shape is not fixed.

So that `crnn` and `segocr` (keep aspect ratio resize) can use batch inference during testing.  
(Note: There may be a mismatch in eval metric for batch_size=1 and batch_size>1)


gaotongxiao (approved):
Batch inference hurts the performance of both models. Users should be clearly warned in the docs.

PR #411, opened 2021-08-04
[Fix] fix single model inference in ocr.py

gaotongxiao:


cuhk-hbsun (approved):

innerlee:
@gaotongxiao could you please add a test case for this in a new pr?

PR #413, opened 2021-08-06
[Docs] Full screen support in readthedocs

gaotongxiao:


gaotongxiao:
This PR can make full use of the screen, not sure if that looks aesthetically fine @innerlee @cuhk-hbsun 
See https://mmocr--413.org.readthedocs.build/en/413/index.html

cuhk-hbsun (approved):

PR #436, opened 2021-08-16
[Refactor] Use MMOCR's registry

gaotongxiao:
# Motivation

The previous implementation of MMOCR registers all the models and layers into MMCV's or MMDetection's registry, leaving the risk of initialization error when two MM-series repos are registering models with the same name into the same registry. This PR solves this issue by creating MMOCR's own registry and introducing model builder helpers that build a model from MMOCR's registry.

# Modification

1. Define MMOCR's registries as children of MMDet's
2. Register all models to MMOCR's own registries
3. Modify some model configs so that some models in MMDet can be correctly located

# BC breaking
If users are using their old configs calling for MMDetection's models, it might not work and has to be modified accordingly. Users should explicitly use `mmdet.` prefix to inform the model builder of the scope to find the model. For example,
```
backbone=dict(
        type='ResNet'
)
```
should be changed to
```
backbone=dict(
        type='mmdet.ResNet'
)
```

# Related PR
https://github.com/open-mmlab/mmsegmentation/pull/790


innerlee:
> Remove some outdated demo scripts

Please remove them in a separate pr. The main change is very complex in itself

cuhk-hbsun (approved):

PR #459, opened 2021-08-24
Support persistent_workers option in data loader

innerlee:
Before this pr, it was ignored.

### DIFF
--- a/tools/test.py
+++ b/tools/test.py
@@ -180,7 +180,6 @@ def main():
                pin_memory=False,
            )),
         **dict((k, cfg.data[k]) for k in [
-                   'samples_per_gpu',
                    'workers_per_gpu',
                    'seed',
                    'prefetch_num',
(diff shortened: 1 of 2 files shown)


gaotongxiao (approved):

tools/test.py
```diff
@@ -180,10 +180,12 @@ def main():
                pin_memory=False,
            )),
         **dict((k, cfg.data[k]) for k in [
+                   'samples_per_gpu',
```
gaotongxiao, line 183: No need to add `samples_per_gpu` here, it will be added back in L195
https://github.com/open-mmlab/mmocr/blob/fba996d07c577d164b29acbd4b2b84870847a33d/tools/test.py#L195

innerlee: I cannot remember why it was special cased at L195...

gaotongxiao: https://github.com/open-mmlab/mmocr/blob/fba996d07c577d164b29acbd4b2b84870847a33d/tools/test.py#L145-L164

gaotongxiao (approved):

Issue #463, opened 2021-08-25
KIE bbox preprocessing may have an error


alexander-soare:
`bboxes` are obtained from `polys` [here](https://github.com/open-mmlab/mmocr/blob/f9d158fafa368e8a5ff9bd41b6c37046df388d08/mmocr/datasets/kie_dataset.py#L215-L226) but I believe there is a fault in the logic.

These lines pick the top-left-most point in the poly, then then 2nd point counter-clockwise from that one.

```
x1, y1 = boxes[:, 0:1], boxes[:, 1:2]  # top-left
x2, y2 = boxes[:, 4:5], boxes[:, 5:6]  # 2nd point counter-clockwise
```

Then later a bounding box is obtained by using these two points:

```
bboxes = np.concatenate([x1, y1, x2, y2], -1).astype(np.float32)
```

BUT, in some cases of non-axis-aligned polygons like this one

![image](https://user-images.githubusercontent.com/16543381/130762188-def66ad2-2673-45cc-9a9a-656f57aa9dd0.png)

the second point is **higher** than the first, resulting in y2 < y1:

![image](https://user-images.githubusercontent.com/16543381/130762437-a10c930a-2ca2-4390-94b0-62645ca9c7b0.png)

Even when this is not the case, it's very possible that the bounding box is not a good representation of what the polygon encapsulates (because it is too narrow).

**I would suggest to instead take the bounding box that contains all points of the poly**. Something like this.

```
bboxes = np.concatenate([polys[:, 0::2].min(axis=1, keepdims=True), polys[:, 1::2].min(axis=1, keepdims=True),
                     polys[:, 0::2].max(axis=1, keepdims=True), polys[:, 1::2].max(axis=1, keepdims=True)], axis=1)
```
Note that this change would also require a change to the logic for the relations calculation (substitute x1, y1, x2, y2 with those from the new `bboxes` def)

Happy to make a PR if it makes sense

gaotongxiao:
You're right, the assumption is oversimplified, tho it's enough to handle the WildReceipt dataset. You're welcome to make a PR :)
BTW, The feature extraction steps and KIE evaluation were designed based on this assumption. You may also check if they still make sense

PR #469, opened 2021-08-28
[Enhancement] Use bounding box around polygon instead of within polygon

alexander-soare:
Address https://github.com/open-mmlab/mmocr/issues/463

Here are the results from training a unet model with and without the modification. (training logs and epoch 60 checkpoints available on request)

Version|Best F1 score|F1 score @ 60 epochs
-|-|-
Old|0.894 @ epoch 43|0.884
New|0.895 @ epoch 44|0.888

The Wildreceipt dataset seems to have mostly axis-aligned polygons so we wouldn't have expected a noticeable different.

### DIFF
--- a/mmocr/datasets/kie_dataset.py
+++ b/mmocr/datasets/kie_dataset.py
@@ -222,6 +222,7 @@ class KIEDataset(BaseDataset):
              boxes[:, 0::2].max(axis=1, keepdims=True),
              boxes[:, 1::2].max(axis=1, keepdims=True)],
             axis=1).astype(np.float32)
+        # yapf: enable
         x1, y1 = bboxes[:, 0:1], bboxes[:, 1:2]
         x2, y2 = bboxes[:, 2:3], bboxes[:, 3:4]
         w, h = np.maximum(x2 - x1 + 1, 1), np.maximum(y2 - y1 + 1, 1)
--- a/mmocr/datasets/kie_dataset.py
+++ b/mmocr/datasets/kie_dataset.py
@@ -215,12 +215,13 @@ class KIEDataset(BaseDataset):
     def compute_relation(self, boxes):
         """Compute relation between every two boxes."""
         # Get minimal axis-aligned bounding boxes for each of the boxes
+        # yapf: disable
         bboxes = np.concatenate(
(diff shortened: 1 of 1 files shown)


alexander-soare:
@gaotongxiao moving our [conversation](https://github.com/open-mmlab/mmocr/issues/463#issuecomment-905351542) here. To answer your last comment:

> The feature extraction steps and KIE evaluation were designed based on this assumption. You may also check if they still make sense

I had a look at these and it seems to me those bits of logic are agnostic to the way the bounding boxes were defined, so long as they are xyxy bounding boxes. Am I missing something? In any case, as you can see in my initial comment, I tested that everything is working fin by training the visual model.

Let me know if I can be helpful with anything else.

alexander-soare:
Looks like there's an issue with linting

My code looks like:

image

But the yapf check is only happy with this

image

Used a `# yapf: disable` as I think my version is far more legible

mmocr/datasets/kie_dataset.py
```diff
@@ -214,13 +214,20 @@ def pad_text_indices(self, text_inds):
...
-        x2, y2 = boxes[:, 4:5], boxes[:, 5:6]
+        # Get minimal axis-aligned bounding boxes for each of the boxes
+        # yapf: disable
+        bboxes = np.concatenate(
+            [boxes[:, 0::2].min(axis=1, keepdims=True),
+             boxes[:, 1::2].min(axis=1, keepdims=True),
+             boxes[:, 0::2].max(axis=1, keepdims=True),
+             boxes[:, 1::2].max(axis=1, keepdims=True)],
+            axis=1).astype(np.float32)
+        x1, y1 = bboxes[:, 0:1], bboxes[:, 1:2]
```
gaotongxiao, line 225:
```suggestion
        # yapf: enable
        x1, y1 = bboxes[:, 0:1], bboxes[:, 1:2]
```
Otherwise yapf would ignore the rest of the code as well

gaotongxiao:
Thanks for your contribution! You're right, everything will be fine as long as we keep using bounding boxes. I mistakenly thought that you were proposing to use the original polygon representation.

@innerlee Any comments?

gaotongxiao (approved):

PR #471, opened 2021-08-30
[Fix]  Allow invalid polygons in intersection and union by default 

### LINKED ISSUE #465: Bad Performance using PANet pre-trained models
Hi there,

I downloaded the PANet checkpoint for CTW and used them to carry out inference on CTW test set. The results that I got are significantly lower than what's reported on [this page](https://github.com/open-mmlab/mmocr/blob/main/configs/textdet/panet/README.md). I have been using all default configurations without any changes. 

The results that I got are:
CTW1500
thr 0.3, recall：0.638, precision: 0.636, hmean:0.637
thr 0.4, recall：0.638, precision: 0.636, hmean:0.637
thr 0.5, recall：0.638, precision: 0.636, hmean:0.637
thr 0.6, recall：0.638, precision: 0.636, hmean:0.637
thr 0.7, recall：0.638, precision: 0.636, hmean:0.637
thr 0.8, recall：0.638, precision: 0.636, hmean:0.637
thr 0.9, recall：0.618, precision: 0.671, hmean:0.643
{'hmean-iou:recall': 0.6182113821138211, 'hmean-iou:precision': 0.6705467372134039, 'hmean-iou:hmean': 0.64331641285956}

I suspect that there is something wrong with the released checkpoints, it seems like it is not trained properly.
Based on this [CTW training log file](https://download.openmmlab.com/mmocr/textdet/panet/panet_r18_fpem_ffm_sbn_600e_ctw1500_20210219-3b3a9aa3.log.json), the validation set's H-mean has already reached 0.66035 at 10th epoch, which is higher than what I get through the released checkpoint.
Could you please help me to verify this?

gaotongxiao:
Fixes #465


mmocr/core/evaluation/utils.py
```diff
@@ -136,24 +136,32 @@ def points2polygon(points):
     return plg(point_mat)
 
 
-def poly_intersection(poly_det, poly_gt, invalid_ret=0, return_poly=False):
+def poly_intersection(poly_det, poly_gt, invalid='allow', return_poly=False):
```
wdsd641417025, line 139: Invalid parameter can be bool type. Using string type need to set scope and assert invalid parameter is set proper

mmocr/core/evaluation/utils.py
```diff
@@ -162,35 +170,39 @@ def poly_intersection(poly_det, poly_gt, invalid_ret=0, return_poly=False):
...
         if return_poly:
-            return invalid_ret, None
+            return invalid, None
         else:
-            return invalid_ret
+            return invalid
 
 
-def poly_union(poly_det, poly_gt, invalid_ret=0, return_poly=False):
+def poly_union(poly_det, poly_gt, invalid='allow', return_poly=False):
```
wdsd641417025, line 178: invalid has multi meanings. Can we use two parameters? We also need check the invalid inputting legality.

mmocr/core/evaluation/utils.py
wdsd641417025, line 139: Invalid has multi meanings. Can we use two parameters? We also need check the invalid inputting legality.

mmocr/core/evaluation/utils.py
wdsd641417025, line 178: same as poly_intersection

mmocr/core/evaluation/utils.py
```diff
@@ -136,23 +136,34 @@ def points2polygon(points):
...
-                                    area. Set as `None` if the input is
-                                    invalid.
+            area. Set as `None` if the input is invalid.
     """
     assert isinstance(poly_det, plg)
     assert isinstance(poly_gt, plg)
+    assert invalid_ret is None or isinstance(invalid_ret, float) or isinstance(
+        invalid_ret, int)
+
+    if invalid_ret is None:
```
wdsd641417025, line 162: Converting invaild poly to poly  can be abstract as a function

mmocr/core/evaluation/utils.py
```diff
@@ -136,23 +136,34 @@ def points2polygon(points):
...
+        if not poly_det.is_valid:
+            poly_det = poly_det.buffer(0)
+        if not poly_gt.is_valid:
+            poly_gt = poly_gt.buffer(0)
 
     if poly_det.is_valid and poly_gt.is_valid:
```
wdsd641417025, line 168: firstly judging return_poly and then is_vaild can  be more  elegant.
```python
poly = None
area = invaild_ret
if return_poly:
    if xx_in_vaild:
        poly = xxx
        area = xxx
    return area, poly
else:
    return area
```

mmocr/core/evaluation/utils.py
```diff
@@ -136,23 +136,34 @@ def points2polygon(points):
     return plg(point_mat)
 
 
-def poly_intersection(poly_det, poly_gt, invalid_ret=0, return_poly=False):
+def poly_intersection(poly_det, poly_gt, invalid_ret=None, return_poly=False):
```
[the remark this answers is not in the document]
gaotongxiao: I don't think so . Your suggestion may be a good practice if we have more than 3 functions implementing similar geometry computations. However, we only implement `union` and `intersection` for now so it is just an over-design.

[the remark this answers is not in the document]
wdsd641417025: You are right and it is may over-design. My suggestion bases on two reason:it can avoid the two places change and it can expend other feature like different backend for intersection and union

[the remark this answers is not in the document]
gaotongxiao: We should keep the implementation simple and make the function name as straightforward as possible. Focusing too much on the extensibilities of two basic utils will just increase the code complexity as of now, given that their implementations are already short, very unlikely to change, and hard to generalize. Also, changing the interface would be a BC-breaking change that should be avoided.

wdsd641417025 (approved):

PR #476, opened 2021-09-01
[Enhancement] Add CITATION.cff

gaotongxiao:


innerlee (approved):

CITATION.cff
```diff
@@ -0,0 +1,9 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
+title: "OpenMMLab Text Detection, Recognition and Understanding Toolbox"
+authors:
+  - name: "MMOCR Contributors"
+version: 0.3.0
```
innerlee, line 6: So this file needs to be updated when release a new version?

gaotongxiao: Right

PR #495, opened 2021-09-16
TextOCR converter: Preserve JPEG quality by copying quantization tables

baudm:
The best way to save the cropped images would be to use PNGs. However, doing so would result in a huge dataset in terms of disk usage. Thus, in order to save the cropped images to as close as an unaltered state as possible (vs original scene images), we keep the quantization tables of the original JPEG images for each of the cropped images.

### DIFF
--- a/tools/data/textrecog/textocr_converter.py
+++ b/tools/data/textrecog/textocr_converter.py
@@ -6,6 +6,7 @@ import os.path as osp
 from functools import partial
 
 import mmcv
+from PIL import Image
 
 from mmocr.utils.fileio import list_to_file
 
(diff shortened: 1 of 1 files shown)


baudm:
Commit 63be435f0e53f3a0348d3edc1016470b80f2ea7f implements a robust orientation correction algorithm for the cropped images. This is done by inferring the general orientation of the text based on the top-left corner of the bounding box and the more fine-grained "bounding points" (`points` array). The first and last indices of the `points` array always correspond to the top-left and bottom-left corners, respectively, of the text (in its own orientation).

gaotongxiao (commented):
Hi, thanks for your contribution! Your work is nice, but we do need a clearer motivation for this PR.
1. I am not familiar with the quantization table thing. Is there any evidence or empirical result showing that preserving the quantization table affects (either positively or negatively) the model's performance?
2. Is it a common practice to correct the cropped images by rotating them? I think it's unfavorable cuz it implicitly changes the data and may result in unfair comparison with other concurrent implementations. Such a trick should have better been implemented as one of the data pipelines that will be explicitly called in the model's config.
3. More motivations and explanations on your rotation trick are preferred. I've got two general questions:
  - How does it deal with vertical texts?
  - (If you believe that your method should definitely be added) Could you illustrate how it works by showing some examples?
4. Your linting doesn't satisfy our requirement. You can use `pre-commit` to fix it.
```
pip install pre-commit
pre-commit install
pre-commit run --files tools/data/textrecog/textocr_converter.py
```

PR #496, opened 2021-09-16
Add converter for the Open Images v5 text annotations by Krylov et al.

baudm:
Open Images V5 Text Annotation and Yet Another Mask Text Spotter
Paper: https://arxiv.org/abs/2106.12326

Similar to TextOCR, this dataset is derived from the Open Images dataset. Particularly, they annotate the `train_1`, `train_2`, `train_5`, `train_f`, and `validation` subsets of the V5 dataset. This results in 2,073,387 word annotations.

Annotations are available here:
https://storage.openvinotoolkit.org/repositories/openvino_training_extensions/datasets/open_images_v5_text

### DIFF
--- a/tools/data/textrecog/openimages_converter.py
+++ b/tools/data/textrecog/openimages_converter.py
@@ -1,0 +1,116 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import argparse
+import math
+import os
+import os.path as osp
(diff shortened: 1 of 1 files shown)


baudm:
Superseded by: https://github.com/open-mmlab/mmocr/pull/497

PR #530, opened 2021-10-19
[Model] Add ABINet vision encoder

gaotongxiao:
# Modifications
1. Add `PositionAttention` into transformer_layers
2. Support `dropout` in `PositionalEncoding`
3. Add `ABIVisionEncoder`
4. Remove unused `self.self_attn` in `TransformerDecoderLayer`
5. Unit tests

### DIFF
--- a/mmocr/models/textrecog/layers/transformer_layer.py
+++ b/mmocr/models/textrecog/layers/transformer_layer.py
@@ -100,7 +100,6 @@ class TransformerDecoderLayer(nn.Module):
                  mask_value=0,
                  act_layer=nn.GELU):
         super().__init__()
-        self.self_attn = MultiHeadAttention()
 
         self.norm1 = nn.LayerNorm(d_model)
         self.norm2 = nn.LayerNorm(d_model)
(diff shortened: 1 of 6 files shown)


mmocr/models/textrecog/layers/transformer_layer.py
```diff
@@ -283,12 +282,143 @@ def forward(self, x):
         return x
 
 
+class PositionAttention(nn.Module):
```
Harold-lkk, line 285: This part can be split two-part: Position and Attention, which are declared in ABIVisionDecoder.

mmocr/models/textrecog/layers/transformer_layer.py
```diff
@@ -283,12 +282,143 @@ def forward(self, x):
...
+
+        self.k_decoder = nn.Sequential(
+            self._decoder_layer(
+                num_channels, num_channels, scale_factor=2, mode=mode),
+            self._decoder_layer(
+                num_channels, num_channels, scale_factor=2, mode=mode),
+            self._decoder_layer(
+                num_channels, num_channels, scale_factor=2, mode=mode),
+            self._decoder_layer(
+                num_channels, in_channels, size=(h, w), mode=mode))
```
Harold-lkk, line 325: This part can be abstract as miniUnet, which is for the domain transfer for key
```suggestion
        self.k = miniUnet(                 
                 in_channels=3,
                 base_channels=num_channels,
                 num_stages=4,
                 strides=((1, 2), (2, 2), (2, 2), (2, 2), (2, 2)),
                 scale_factor=(2,2,2,None),
                 size=(None, None, None, (h, w)),
                 mode=(mode, mode, mode, mode)
                 )
```
reference the implement in mmocr'unet

gaotongxiao: The structure is very simple and adding it to backbones might be unnecessary. We can leave it as it is until some other tasks have used a similar structure.

Harold-lkk: It does not need to add it to backbones. Actually, it is just a nonlinear transform from feature to key. But written all in the class, it's difficult to understand it immediately， while abstracting it as `self.k_project = miniUnet()`, it's no need to read miniUnet for knowing this function. Also, essentially, PositionAttention is a standard Attention, except q is PositionalEncoding, which should be passed to attention, and key is projected with a complex net.

gaotongxiao: I still don't think we should add mini-unet.
1. In `forward()`, there are sufficient comments and clearly divided code blocks for users to understand what these layers are doing.
2. I think we are mostly sure that it won't be reused by other network implementations. 
3. The suggested interface is overly complicated and increases the cost of understanding for users. Imagine that a user has to jump between files to understand how this network has been created, then they realize that it can be clearly presented within 8 lines. 

Overall, I think the **reusability** matters the most - if we don't believe that this sub-network can significantly benefit other models and therefore it is worth being more generalized, we can save efforts on separating it out from the current module. Instead, we just make sure comments are enough to keep users from being lost.

Harold-lkk: The decode used transformer in MMOCR,  q = nonlinear(feat)(position), k = nonlinear(feat)(complex | optional) v = feat and DotProductAttentionLayer(q, k, v mask)

Harold-lkk: if it won't be reused by other network implementations, it may merge the PositionAttention into ABIVisionDecoder directly, not building a PositionAttention class individually

PR #577, opened 2021-11-11
[Fix] Fix resource warning: unclosed file

cuhk-hbsun:


gaotongxiao (commented):
Just noticed that det_test_imgs.py should be also fixed as #574

gaotongxiao (approved):

PR #581, opened 2021-11-12
[Fix] Modify algorithm "sar" weights path in metafile

ShoupingShan:
Original model `sar_r31_parallel_decoder_academic` weights path in sar/metafile.yml belongs to crnn which should be modified.


gaotongxiao (approved):
Thanks! Please sign CLA before we are able to merge your PR.

PR #603, opened 2021-11-19
[Enhancement] Remove unnecessary custom_import from train.py

gaotongxiao:
Removing it as MMCV can handle this field.

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -111,10 +111,6 @@ def main():
             cfg.data.train.pipeline[0].update(
                 file_client_args=mc['mc_file_client_args'])
 
-    # import modules from string list.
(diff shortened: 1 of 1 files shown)


Harold-lkk:
so where to import custom import

gaotongxiao:
@zhouzaida

zhouzaida (approved):

Harold-lkk:
Do we need to limit MMCV version?

gaotongxiao:
MMCV started to support this from 1.1.6.

cuhk-hbsun (approved):

Harold-lkk:
Reference to https://github.com/open-mmlab/mmcv/pull/606

PR #652, opened 2021-12-07
[Enhancement] Use shape from loaded image

cuhk-hbsun:
Use the shape of the loaded image so that fake `width` and `height` can be given in the annotation file.

### DIFF
--- a/mmocr/datasets/pipelines/loading.py
+++ b/mmocr/datasets/pipelines/loading.py
@@ -42,6 +42,10 @@ class LoadTextAnnotations(LoadAnnotations):
     def _load_masks(self, results):
         ann_info = results['ann_info']
         h, w = results['img_info']['height'], results['img_info']['width']
+        if results.get('ori_shape', None):
+            h, w = results['ori_shape'][:2]
+            results['img_info']['height'] = h
+            results['img_info']['width'] = w
         gt_masks = ann_info['masks']
         if self.poly2mask:
             gt_masks = BitmapMasks(


tests/test_dataset/test_loading.py
```diff
@@ -31,12 +31,26 @@ def test_loadtextannotation():
     with_seg = False
     poly2mask = False
 
-    loader = LoadTextAnnotations(with_bbox, with_label, with_mask, with_seg,
-                                 poly2mask)
+    loader = LoadTextAnnotations(
```
gaotongxiao, line 34: Add a comment explaining what's being tested here

tests/test_dataset/test_loading.py
```diff
@@ -31,12 +31,26 @@ def test_loadtextannotation():
...
+        with_label,
+        with_mask,
+        with_seg,
+        poly2mask,
+        use_img_shape=True)
     output = loader._load_masks(results)
     assert len(output['gt_masks_ignore']) == 4
     assert np.allclose(output['gt_masks_ignore'].masks[0],
                        [[499, 94, 531, 94, 531, 124, 499, 124]])
+    loader = LoadTextAnnotations(
```
gaotongxiao, line 45: Same

tests/test_dataset/test_loading.py
```diff
@@ -31,12 +31,31 @@ def test_loadtextannotation():
...
+        with_mask,
+        with_seg,
+        poly2mask,
+        use_img_shape=True)
     output = loader._load_masks(results)
     assert len(output['gt_masks_ignore']) == 4
     assert np.allclose(output['gt_masks_ignore'].masks[0],
                        [[499, 94, 531, 94, 531, 124, 499, 124]])
+    # If 'ori_shape' in result and use_img_shape=True,
+    # result['ori_shape'] will be used to generate mask.
```
gaotongxiao, line 49: But there is no assertion to validate the result?

gaotongxiao (approved):

PR #657, opened 2021-12-09
[Enhancement] Support muitiple pipelines for different datasets

cuhk-hbsun:
For example, It is useful when datasets are stored in different devices (hard-disk, aws-ceph). In that way, different `pipelines` should be applied to different datasets.

### DIFF
--- a/mmocr/apis/utils.py
+++ b/mmocr/apis/utils.py
@@ -1,4 +1,5 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+import copy
 import warnings
 
 from mmdet.datasets import replace_ImageToTensor
@@ -19,6 +20,8 @@ def replace_image_to_tensor(cfg, set_types=None):
     assert set_types is None or isinstance(set_types, list)
     if set_types is None:
         set_types = ['val', 'test']
+
+    cfg = copy.deepcopy(cfg)
     for set_type in set_types:
         assert set_type in ['val', 'test']
         uniform_pipeline = cfg.data[set_type].get('pipeline', None)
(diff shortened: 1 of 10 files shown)


mmocr/apis/utils.py
```diff
@@ -0,0 +1,86 @@
...
+            warnings.warn(warning_msg)
+            cfg.pipeline = [cfg.pipeline[0], *cfg.pipeline[1].transforms]
+    else:
+        if cfg[idx][1].type == 'MultiRotateAugOCR':
+            warnings.warn(warning_msg)
+            cfg[idx] = [cfg[idx][0], *cfg[idx][1].transforms]
+
+
+def disable_text_recog_aug_test(cfg, set_types=None):
+    """Remove aug_test from test pipeline for text recognition.
```
innerlee, line 54: In-place modification is dangerous since the original cfg will be modified and there is no `enable_xxx` to revert the change.

mmocr/datasets/uniform_concat_dataset.py
```diff
@@ -3,27 +3,65 @@
...
+                 **kwargs):
+        new_datasets = []
         if pipeline is not None:
-            assert from_cfg, 'datasets should be config dicts'
-            for dataset in datasets:
-                if dataset['pipeline'] is None:
-                    dataset['pipeline'] = copy.deepcopy(pipeline)
-        datasets = [build_dataset(c, kwargs) for c in datasets]
+            assert len(
+                pipeline
```
innerlee, line 39: a string has length > 0

innerlee (approved):

PR #663, opened 2021-12-10
[FIX] #665 fix change locals() dict

### LINKED ISSUE #665: locals() should not be modified
```python
args = locals()
[args.pop(x, None) for x in ['kwargs', 'self']]
```
https://github.com/open-mmlab/mmocr/blob/b04775fd78ac89e32d38ef6fbd5493dedbfd76f4/mmocr/utils/ocr.py#L414

Fei-Wang:
fix #665 

locals() dict should not be changed. [ref](https://docs.python.org/3/library/functions.html#locals)

If locals() runs more than once or during breakpoint debugging, the value of args will automatically become the dictionary before modification. 

In this case, a shallow copy is better.

### DIFF
--- a/mmocr/utils/ocr.py
+++ b/mmocr/utils/ocr.py
@@ -411,7 +411,7 @@ class MMOCR:
                  merge=False,
                  merge_xdist=20,
                  **kwargs):
-        args = locals()
+        args = locals().copy()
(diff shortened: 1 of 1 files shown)


gaotongxiao:
It's not even a bug in the current context, but it's great to maintain a good practice across our implementations. I'm merging it without explicit testing.

PR #675, opened 2021-12-15
[Fix] Support both ConcatDataset and UniformConcatDataset

cuhk-hbsun:


gaotongxiao (approved):
LGTM. @AllentDan Could you double-check if the changes in deployment scripts would break any part on your side?

AllentDan:
OK for my side

PR #685, opened 2021-12-23
[Fix] Use RECOGNIZER instead of DETECTORS

### LINKED ISSUE #595: RECOGNIZERS Registry Code Design
I find the Registry RECOGNIZERS defined in mmocr/models/builder.py never be used in somewhere else, and the "recognizer" model like SARNet was registered in DETECTOR Register. Is here some reason about this code design?

cuhk-hbsun:
fix #595


mmocr/models/builder.py
```diff
@@ -22,6 +22,7 @@
 HEADS = BACKBONES
 NECKS = BACKBONES
 FUSERS = BACKBONES
+RECOGNIZERS = BACKBONES
```
gaotongxiao, line 25: Remove line 10

gaotongxiao:
BTW, it's not a breaking change since RECOGNIZERS is an alias of `model` Registry @Harold-lkk

Harold-lkk:
@gaotongxiao  actually it is. Using RECOGNIZERS need to build the model with build_recognizer,  which effect a lot, like train, test, inference, torchserve and depoly

gaotongxiao:
@Harold-lkk `build_detector` still works implicitly. Its name may arise some confusion though. @cuhk-hbsun may need to rename `init_detector`, `build_detector` to more general names like `init_models` and``build_models`.

Harold-lkk:
build_detector will not work, because ` DETECTORS.build(
        cfg, default_args=dict(train_cfg=train_cfg, test_cfg=test_cfg))`. Also if use build model, need to get a task name, which actually effect a lot which may need a doc for designing

gaotongxiao:
@Harold-lkk I didn't get your point. Anyway, recognizers have passed the test build.
https://github.com/open-mmlab/mmocr/blob/0a1787d6bcbec25649441a1f26e928ea7158e4c4/tests/test_models/test_recog_config.py#L108-L119

gaotongxiao:
Need a follow-up PR that renames init_detector before we merge this one. Otherwise, it would become rather confusing.

gaotongxiao (approved):
Need a follow-up PR that renames init_detector before we merge this one. Otherwise, it would become rather confusing.

PR #716, opened 2022-01-06
Update sar_r31_parallel_decoder_chinese.py

TommyZihao:
fix the bug of Chinese OCR


cuhk-hbsun:
Thanks for your contribution but why it is a bug if you put the `dict_file` under `data/chineseocr/labels`? Is anything wrong?

PR #728, opened 2022-01-12
Extend totaltext converter to support text fields

Harold-lkk:
## Motivation

Extend totaltext converter to support text fields

## Modification

Add text in annotation.

### DIFF
--- a/tools/data/textdet/totaltext_converter.py
+++ b/tools/data/textdet/totaltext_converter.py
@@ -241,7 +241,7 @@ def get_contours_txt(gt_path):
         if tmp_line != '':
             contours, words = process_line(tmp_line, contours, words)
 
-        words = [word.replace('#', '###') for word in words]
+        words = ['###' if word == '#' else word for word in words]
 
     return contours, words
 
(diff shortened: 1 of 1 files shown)


tools/data/textdet/totaltext_converter.py
```diff
@@ -239,11 +241,7 @@ def get_contours_txt(gt_path):
         if tmp_line != '':
             contours, words = process_line(tmp_line, contours, words)
 
-        for word in words:
-
-            if word == '#':
-                word = '###'
-                continue
+        words = [word.replace('#', '###') for word in words]
```
gaotongxiao, line 244: Not rigorous enough. What if the word itself is '###'?

tools/data/textdet/totaltext_converter.py
```diff
@@ -87,7 +87,8 @@ def get_contours_mat(gt_path):
     contours = []
     words = []
     data = scio.loadmat(gt_path)
-    data_polygt = data['polygt']
+    # 'gt' in latest version; 'polygt' in old version
```
gaotongxiao, line 90:
```suggestion
    # 'gt' for the latest version; 'polygt' for the legacy version
```

gaotongxiao (approved):

PR #737, opened 2022-01-17
[Feature] Add GenerateBezierPoints

yuexy:


mmocr/datasets/pipelines/spotting_transforms.py
```diff
@@ -0,0 +1,73 @@
...
+        assert check_argument.is_2dlist(polygon_masks)
+
+        bezier_points = []
+        for poly in polygon_masks:
+            poly = poly[0]
+            assert len(poly) % 4 == 0
+            num_points = int(len(poly) / 2)
+            num_side_points = int(len(poly) / 4)
+
+            curve_data_top = poly[0:num_points].reshape(num_side_points, 2)
```
gaotongxiao, line 52:
```suggestion
            curve_data_top = poly[:num_points].reshape(num_side_points, 2)
```

mmocr/datasets/pipelines/spotting_transforms.py
```diff
@@ -0,0 +1,73 @@
...
+    def __call__(self, results):
+        polygon_masks = results['gt_masks'].masks
+        assert check_argument.is_2dlist(polygon_masks)
+
+        bezier_points = []
+        for poly in polygon_masks:
+            poly = poly[0]
+            assert len(poly) % 4 == 0
+            num_points = int(len(poly) / 2)
+            num_side_points = int(len(poly) / 4)
```
gaotongxiao, line 50: Two assumptions here: 1. there must be 2k points in the polygonmask; 2. the first and last k points forms side curves. I'm concerned about its generalizability: Is this transform applicable to *all* kinds of point annotations? If not, shall we provide util to format the points in this way?

PR #756, opened 2022-01-26
[Feat] Support specify gpu  for testing and training with gpu-id  instead of gpu-ids and gpus 

Harold-lkk:
https://github.com/open-mmlab/mmdetection/pull/7019
## Motivation
1. support test for specifying GPU
2. Since MMDP does not support non-distributed multi-GPU training, --gpus in train.py lost its role, so it is removed, --gpu-ids is changed to --gpu-id, because only one GPU can be specified for non-distribution training and testing.
If the number of GPUs is more than 1, there will be an assertion error in MMDP because MMDP does not support the situation with more than one GPU.

## Modification
--gpu-ids is changed to --gpu-id.

## BC-breaking (Optional)
--gpus-ids and --gpus will be deprecated, please use --gpu-id.


gaotongxiao (commented):
Need to update docs (training.md & test.md)

tools/train.py
```diff
@@ -128,9 +134,13 @@ def main():
     if args.resume_from is not None:
         cfg.resume_from = args.resume_from
     if args.gpu_ids is not None:
```
gaotongxiao, line 136: missing https://github.com/open-mmlab/mmdetection/blob/277e5dc48b67a87d51e7636966e555eb52196471/tools/train.py#L118-L122 ?

docs/en/training.md
```diff
@@ -12,20 +12,21 @@ python tools/train.py ${CONFIG_FILE} [ARGS]
...
+| `--gpu-ids`       | int*N                             | Deprecated, please use --gpu-id. A list of GPU ids to use. Only applicable to non-distributed training.                                                                                                                                                                                                                                                                                |
+| `--gpu-id`        | int                               | GPU id to use. Only applicable to non-distributed training.                                                                                                                                                                                                                                                                                                                            |
```
gaotongxiao, line 23:
```suggestion
| `--gpus`          | int                               | **Deprecated, please use --gpu-id.** Numbers of GPUs to use. Only applicable to non-distributed training.                                                                                                                                                                                                                                                                                   |
| `--gpu-ids`       | int*N                             | **Deprecated, please use --gpu-id.** A list of GPU ids to use. Only applicable to non-distributed training.                                                                                                                                                                                                                                                                                |
| `--gpu-id`        | int                               | The GPU id to use. Only applicable to non-distributed training.                                                                                                                                                                                                                                                                                                                            |
```

gaotongxiao (approved):

PR #776, opened 2022-02-09
[Model] FCOS detector

Harold-lkk:
- [x] config
- [x] FCOS Head, Loss
- [x] BaseTextDetPostprocessor, FCOSPostprocessor
- [x] Redesigning dataset evaluator
- [x] unit tests


gaotongxiao:
Status update: FCOS is confirmed to be equivalent to that in mmdet, but `cls_loss` explodes after a few training iterations. Still unclear about its reason.

PR #801, opened 2022-02-28
Add deprecation message for deploy tools

xinke-wang:
## Motivation

Since [MMDeploy](https://github.com/open-mmlab/mmdeploy) has been released, the deployment tools scattered in each OpenMMLab codebase will be deprecated in the future.

## Modification

Add deprecation message in deployment tools.

## Use cases (Optional)

image

### DIFF
--- a/tools/deployment/deploy_test.py
+++ b/tools/deployment/deploy_test.py
@@ -46,6 +46,19 @@ def parse_args():
 
 def main():
     args = parse_args()
+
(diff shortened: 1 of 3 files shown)


xinke-wang:
> Deprecation message should be placed somewhere after `parse_args()` so that users can see it when the script starts.

Done

gaotongxiao (approved):

Issue #816, opened 2022-03-07
can't start training using multiple GPUs for MMOCR sar Model with custom Dataset


payal211:
Hi @gaotongxiao 

I am training sar model with custom dataset on AWS with 4 GPUs of 24 GB.

I started training using below command:
CUDA_VISIBLE_DEVICES=0,1,2,3 GPUS=4 python tools/train.py configs/textrecog/sar/sar_r31_parallel_decoder_own_dataset.py --work-dir sar

training is started and it shows Approx. 495 days to complete.
Training SS:
![Training_ss](https://user-images.githubusercontent.com/43141616/157045336-3fe05f2d-4b0d-4d75-9570-5636e61f8af3.png)

and it is not using multiple GPUs. here is the info of nvidia-smi command:
![nvidia-output](https://user-images.githubusercontent.com/43141616/157045858-a2694eab-3b3d-4231-a627-67439dbf1899.png)

Can you please Help me out for this problem. and guide me how to start training using multiple GPUs to decrease training time.

Thanks.

xinke-wang:
Hi, try to use the following command:

`./tools/dist_train.sh configs/textrecog/sar/sar_r31_parallel_decoder_own_dataset.py  sar 4`

You can also read the **Training on Multiple Machines** section in the [MMOCR doc](https://mmocr.readthedocs.io/en/latest/training.html).

gaotongxiao:
Sorry for causing the confusion, the documentation is inaccurate. Directly calling `tools/train.py` uses one GPU. To use multiple GPUs, please follow what @xinke-wang said. We need to fix this section to avoid further confusion.

PR #819, opened 2022-03-09
[Docs] Correct misleading section title in training.md

gaotongxiao:
## Motivation

The only way to use multiple GPUs for training is via `dist_train.sh`, whereas our docs can be misleading. Addresses #816


xinke-wang (approved):
LGTM

PR #854, opened 2022-03-18
[Fix] Support loading space character from dict file

gaotongxiao:

## Motivation

`BaseConvertor` used to skip any whitespace when loading a dictionary from a file. For example, one passes a legal dictionary file to `label_convertor` where the third line is a space character:

```
a
b
 
d
```

And they would expect the label convertor to create the dictionary as `['a', 'b', ' ', 'd']` but get `['a', 'b', 'd']` instead. This PR improves the parser by stripping only line-breaking characters (`\r` and `\n`) from each line and retaining the space character. 

## Modifications

1. Improve on the parsing part of `BaseConvertor.__init__()`
3. Add more unit tests for `BaseConvertor`

## BC-breaking (Optional)

A dictionary containing duplicated characters will be regarded as an invalid input and break the assertion of the label convertor. It was permitted implicitly before, but usually we don't want `label_convertor` to generate ambiguous results.

### DIFF
--- a/mmocr/models/textrecog/convertors/base.py
+++ b/mmocr/models/textrecog/convertors/base.py
@@ -24,26 +24,33 @@ class BaseConvertor:
                    '*+,-./:;<=>?@[\\]_`~')
 
     def __init__(self, dict_type='DICT90', dict_file=None, dict_list=None):
-        assert dict_type in ('DICT36', 'DICT90')
         assert dict_file is None or isinstance(dict_file, str)
         assert dict_list is None or isinstance(dict_list, list)
         self.idx2char = []
         if dict_file is not None:
-            for line in list_from_file(dict_file):
-                line = line.strip()
+            for line_num, line in enumerate(list_from_file(dict_file)):
+                line = line.strip('\r\n')
+                if len(line) > 1:
+                    raise ValueError('Expect each line has 0 or 1 character, '
+                                     f'got {len(line)} characters '
+                                     f'at line {line_num + 1}')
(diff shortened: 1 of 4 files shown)


xinke-wang:
I have a question is that the 'DICT36' and 'DICT90' do not contain space. So, if the users would like to train a dataset that contains space, they have to modify the DICT manually. Do we need to provide an optional parameter (or a dict containing space) to allow users to enable space recognition in a faster way?

gaotongxiao:
@xinke-wang Space character is not as special as `<UNK>`, `<PAD>`, `<BOS/EOS>`, so principally we should not design an optional argument for space recognition. Providing `DICT37` and `DICT91` and informing users of their availability in both the tutorial (https://github.com/open-mmlab/mmocr/pull/850) and the release log sound more appealing to me.

xinke-wang:
> @xinke-wang Space character is not as special as `<UNK>`, `<PAD>`, `<BOS/EOS>`, so principally we should not design an optional argument for space recognition. Providing `DICT37` and `DICT91` and informing users of their availability in both the tutorial (#850) and the release log sound more appealing to me.

Ok, I will add instructions to #850.

xinke-wang (approved):

xinke-wang:
Since we now support `DICT37` and `DICT91`, the docstring of BaseConvertor should be updated.

`dict_type (str): Type of dict, should be either 'DICT36' or 'DICT90'`

xinke-wang (approved):

PR #869, opened 2022-03-23
[Fix] ignore_index in SARLoss

Mountchicken:
## Motivation
When caculating `CrossEntropy `loss in attention decoder, we can assign `ignore_index` to skip the loss of `<PAD>` token. In MMOCR, this assigning process is automatically done at [here](https://github.com/open-mmlab/mmocr/blob/main/mmocr/models/textrecog/recognizer/encode_decode_recognizer.py#L60)
```python
loss.update(ignore_index=self.label_convertor.padding_idx)
```
However, in [SARLoss](https://github.com/open-mmlab/mmocr/blob/main/mmocr/models/textrecog/losses/ce_loss.py#L67), the ignore index is initialized as `0` which may cause confusing, even if it won't work as we can automatically update the `ignore_index`. So we'd like to reset this value to `-1`.

### DIFF
--- a/mmocr/models/textrecog/losses/ce_loss.py
+++ b/mmocr/models/textrecog/losses/ce_loss.py
@@ -79,7 +79,7 @@ class SARLoss(CELoss):
         SARLoss assumes that the first input token is always `<SOS>`.
     """
 
-    def __init__(self, ignore_index=0, reduction='mean', **kwargs):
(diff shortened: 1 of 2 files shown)


gaotongxiao (commented):
Two fixes should be separated into two PRs.

gaotongxiao (approved):

cuhk-hbsun (approved):

PR #872, opened 2022-03-23
[Feature] Add COCO Text v2 Data Converter

xinke-wang:
## Motivation

Add data converter for [COCO Text v2](https://bgshih.github.io/cocotext/).

## Modification

add `tools/data/textdet/cocotext_converter.py` and `tools/data/textrecog/cocotext_converter.py`

### COCO Text v2

- Step1: Download image [train2014.zip](http://images.cocodataset.org/zips/train2014.zip) and annotation [cocotext.v2.zip](https://github.com/bgshih/cocotext/releases/download/dl/cocotext.v2.zip) to `coco_textv2/`.

```bash
mkdir coco_textv2 && cd coco_textv2
mkdir annotations

# Download COCO Text v2 dataset
wget http://images.cocodataset.org/zips/train2014.zip
wget https://github.com/bgshih/cocotext/releases/download/dl/cocotext.v2.zip
unzip -q train2014.zip && unzip -q cocotext.v2.zip

mv train2014 imgs && mv cocotext.v2.json annotations/

rm train2014.zip && rm -rf cocotext.v2.zip
```

## Detection

- Step2: Generate `instances_training.json` and `instances_val.json` with the following command:

```bash
python tools/data/textdet/cocotext_converter.py PATH/TO/coco_textv2
```

## Recognition

- Step2: Generate `train_label.jsonl` and `val_label.jsonl` with the following command:

```bash
python tools/data/textrecog/cocotext_converter.py PATH/TO/coco_textv2 --nproc 4
```


docs/en/datasets/recog.md
```diff
@@ -79,26 +79,31 @@
 │   │   ├── annotations
 │   │   ├── train_label.txt
 │   │   ├── test_label.txt
+│   ├── coco_textv2
+│   │   ├── crops
+│   │   ├── train_label.jsonl
+│   │   ├── val_label.jsonl
```
Mountchicken, line 85: There is one thing that confuses me. No matter we are running recognition or detection script, files of `imgs` and `annotations` will always exist. But we don't need them in recognition. So, should we list their names here?

xinke-wang: > There is one thing that confuses me. No matter we are running recognition or detection script, files of `imgs` and `annotations` will always exist. But we don't need them in recognition. So, should we list their names here?

That's also my concern. I only list the related dirs for the certain task (like only cropped imgs & labels for recog, while imgs, annotation for det). What do you @gaotongxiao think?

gaotongxiao: What @xinke-wang did is right. We don't need to list out all the files that are irrelevant to this task.

Mountchicken (approved):

PR #908, opened 2022-04-02
[Docs] Fix the misleading description in test.py

gaotongxiao:


xinke-wang:
We may also need to update the doc

xinke-wang (approved):

PR #922, opened 2022-04-11
openset sdmgr metakey correct

garvan2021:
the original meta key `ori_boxes` will enconter ``KeyError`` at inference ocr.py https://github.com/open-mmlab/mmocr/blob/a3834ab24e47f951138e1a8a6ca396e36eea6e2e/mmocr/utils/ocr.py#L742


gaotongxiao:
@garvan2021 Sorry for the confusion. It happens because there is a mismatch between your local git email address and your actual GitHub account. Even though you have signed CLA with your github account, CLA assistant is still not able to link this address to your commit.

Try to correct your email address locally and amend your commits again:
```bash
git config --global user.email "youremail@yourdomain.com"
```

PR #926, opened 2022-04-12
correct meta key

garvan2021:


garvan2021:
extend from https://github.com/open-mmlab/mmocr/pull/922

cuhk-hbsun (approved):

gaotongxiao (commented):
Hi @garvan2021 , there are two other places needed to be fixed as well:
https://github.com/open-mmlab/mmocr/blob/a50b0c9fb9196fbd5208400a32ec0bbc068d3e8b/tests/test_dataset/test_openset_kie_dataset.py#L87
https://github.com/open-mmlab/mmocr/blob/a50b0c9fb9196fbd5208400a32ec0bbc068d3e8b/tests/test_dataset/test_openset_kie_dataset.py#L92

(though these keys won't affect the test results)

gaotongxiao (approved):

PR #933, opened 2022-04-13
[Enhancement] More customizable fields in dataloaders

gaotongxiao:
## Motivation

This PR enhances the configuration flexibility of `train_dataloader`, `val_dataloader` and `test_dataloader`. Users can customize more options in dataloaders rather than a subset of predefined fields. This PR also refines default dataloader configs.

## Relevant PR
https://github.com/open-mmlab/mmdetection/pull/7668

### DIFF
--- a/mmocr/apis/train.py
+++ b/mmocr/apis/train.py
@@ -30,12 +30,18 @@ def train_detector(model,
     # prepare data loaders
     dataset = dataset if isinstance(dataset, (list, tuple)) else [dataset]
     # step 1: give default values and override (if exist) from cfg.data
-    default_loader_cfg = dict(
-        num_gpus=len(cfg.gpu_ids),
-        dist=distributed,
-        seed=cfg.get('seed'),
-        drop_last=False,
-        persistent_workers=False)
+    default_loader_cfg = {
+        **dict(
+            num_gpus=len(cfg.gpu_ids),
+            dist=distributed,
+            seed=cfg.get('seed'),
+            drop_last=False,
+            persistent_workers=False),
+        **({} if torch.__version__ != 'parrots' else dict(
+               prefetch_num=2,
+               pin_memory=False,
+           )),
+    }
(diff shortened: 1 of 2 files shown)


mmocr/apis/train.py
```diff
@@ -30,30 +30,31 @@ def train_detector(model,
...
+        for k, v in cfg.data.items() if k not in [
+            'train', 'val', 'test', 'train_dataloader', 'val_dataloader',
+            'test_dataloader'
+        ]
+    })
+    train_loader_cfg = {
+        **default_loader_cfg,
         **({} if torch.__version__ != 'parrots' else dict(
                prefetch_num=2,
                pin_memory=False,
```
Mountchicken, line 51: This configuration was previously available in the val_dataloader, but now it is only available in the train_dataloader. Should we move them to default_lodaer_cfg?

Mountchicken:
BTW, should we modify our entire config file as mmdet did, to better explain the current dataloader configuration method?

gaotongxiao:
@Mountchicken The way how MMOCR parses the config remains the same after this PR, so the configs don't need to be migrated   and commented

Mountchicken:
@gaotongxiao Get your mean, but it feels like it would be clearer for users who don't understand the structure our config file. Of course, the core of this problem is that we should explain our config file in more details.
```python
# current one
data = dict(
    samples_per_gpu=64,
    workers_per_gpu=4,
    val_dataloader=dict(samples_per_gpu=1),
    test_dataloader=dict(samples_per_gpu=1),
# suggest one
data = dict(
    train_dataloader-dict( samples_per_gpu=64, workers_per_gpu=4),
    val_dataloader=dict(samples_per_gpu=1, workers_per_gpu=4),
    test_dataloader=dict(samples_per_gpu=1, workers_per_gpu=4),
```

Mountchicken (approved):

PR #949, opened 2022-04-20
[Fix] Fix ImgAug transform

gaotongxiao:
## Motivation

The current implementation of ImgAug transform transforms bounding box points [x1, y1, x2, y2] in the same way as keypoints, and the resulting keypoints [x3, y3, x4, y4] can break the basic assumptions in bounding boxes that (x4 > x3) and (y4 > y4). This PR transforms the bounding box in the right way to make sure the resulting box still bounds the instance region. It also handles the out-of-boundary bounding boxes by clipping out the invalid regions.

The same issue also exists in polygon transform. Treating a polygon as a collection of keypoints way too simplifies the assumption. The resulting polygons without boundary tests can cross the boundary and become noise data. This PR help clip the illegal regions as well.

Using this augmentation to train DBNet can slightly increase its performance.

### DIFF
--- a/mmocr/datasets/pipelines/dbnet_transforms.py
+++ b/mmocr/datasets/pipelines/dbnet_transforms.py
@@ -57,11 +57,14 @@ class ImgAug:
             followed by random rotation with angles in range [-10, 10], and
             resize with an independent scale in range [0.5, 3.0] for each
             side of images.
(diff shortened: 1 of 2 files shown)


xinke-wang (approved):
Works good on dbnet and dbnet++

PR #951, opened 2022-04-20
[Feature] Add EAST head

Mountchicken:
- Modules
- [x] EAST neck
- [x] EAST head
- [ ] EAST loss
- [ ] EAST postprocessor
- [ ] EAST Targets
- Unit Tests
- [x] EAST neck
- [ ] EAST head
- [ ] EAST loss
- [ ] EAST postprocessor
- [ ] EAST Targets

### DIFF
--- a/mmocr/models/textdet/dense_heads/east_head.py
+++ b/mmocr/models/textdet/dense_heads/east_head.py
@@ -1,0 +1,88 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from mmcv.cnn import ConvModule
+from mmcv.runner import BaseModule, Sequential
+
+from mmocr.models.builder import HEADS
+from .head_mixin import HeadMixin
+
+
+@HEADS.register_module()
+class EASTHead(HeadMixin, BaseModule):
+
+    def __init__(self,
+                 box_type,
+                 num_classes,
+                 in_channels,
+                 loss=dict(type='EASTLoss', box_type='RBOX'),
+                 postprocessor=dict(
+                     type='EASTPostprocessor',
+                     text_repr_type='quad',
+                     box_type='RBOX'),
+                 init_cfg=[
+                     dict(type='Kaiming', layer='Conv'),
+                     dict(
+                         type='Constant', layer='BatchNorm', val=1., bias=1e-4)
+                 ],
+                 train_cfg=None,
+                 test_cfg=None):
+        BaseModule.__init__(self, init_cfg=init_cfg)
+        HeadMixin.__init__(self, loss, postprocessor)
+        assert box_type in ['RBOX', 'QUAD']
+        assert box_type == loss['box_type'] == postprocessor['box_type']
+        self.box_type = box_type
+        self.num_classes = num_classes
+        self.in_channels = in_channels
+        self.train_cfg = train_cfg
+        self.test_cfg = test_cfg
+        self._init_layers()
+
+    def _init_layers(self):
+        """Initialize layers of the head."""
+        self._init_cls_convs()
+        self._init_reg_convs()
+
+    def _init_reg_convs(self):
+        """Initialize bbox regression conv layers of the head."""
+        if self.box_type == 'RBOX':
+            self.ang_conv = Sequential(
+                ConvModule(
+                    self.in_channels,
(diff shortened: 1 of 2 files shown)


mmocr/models/textdet/dense_heads/east_head.py
```diff
@@ -0,0 +1,88 @@
...
+        self.num_classes = num_classes
+        self.in_channels = in_channels
+        self.train_cfg = train_cfg
+        self.test_cfg = test_cfg
+        self._init_layers()
+
+    def _init_layers(self):
+        """Initialize layers of the head."""
+        self._init_cls_convs()
+        self._init_reg_convs()
```
xinke-wang, line 42: These `_init_layers`, `_init_cls_convs`, and `_init_reg_convs` look a little bit weird here. Is it necessary to split these definitions into two functions? When I look at the names, my first impression is that these functions are used for initializing weights. Directly defining `self.score_conv` and `self.rec_conv` in `__init__` might be enough

mmocr/models/textdet/dense_heads/east_head.py
```diff
@@ -0,0 +1,88 @@
...
+    def _init_cls_convs(self):
+        """Initialize score prediction conv layers of the head."""
+        self.score_conv = Sequential(
+            ConvModule(
+                self.in_channels,
+                out_channels=1,
+                kernel_size=1,
+                stride=1,
+                norm_cfg=dict(type='BN'),
+                act_cfg=dict(type='ReLU')))
```
xinke-wang, line 74: The [official implementation](https://github.com/argman/EAST/blob/master/model.py#L74) uses `Sigmoid` activation in all output layers.
```suggestion
                act_cfg=dict(type='Sigmoid')))
```

mmocr/models/textdet/dense_heads/east_head.py
```diff
@@ -0,0 +1,88 @@
...
+        """Initialize bbox regression conv layers of the head."""
+        if self.box_type == 'RBOX':
+            self.ang_conv = Sequential(
+                ConvModule(
+                    self.in_channels,
+                    out_channels=1,
+                    kernel_size=1,
+                    stride=1,
+                    norm_cfg=dict(type='BN'),
+                    act_cfg=dict(type='ReLU')))
```
xinke-wang, line 54: The [official implementation](https://github.com/argman/EAST/blob/master/model.py#L74) uses `Sigmoid` activation in all output layers.
```suggestion
                    act_cfg=dict(type='Sigmoid')))
```

mmocr/models/textdet/dense_heads/east_head.py
```diff
@@ -0,0 +1,88 @@
...
+
+        out_channels = 4 if self.box_type == 'RBOX' else 8
+        self.reg_conv = ConvModule(
+            self.in_channels,
+            out_channels,
+            kernel_size=1,
+            stride=1,
+            norm_cfg=dict(type='BN'),
+            act_cfg=dict(type='ReLU'))
```
xinke-wang, line 63: The [official implementation](https://github.com/argman/EAST/blob/master/model.py#L74) uses `Sigmoid` activation in all output layers.
```suggestion
            act_cfg=dict(type='Sigmoid'))
```

mmocr/models/textdet/dense_heads/east_head.py
```diff
@@ -0,0 +1,88 @@
...
+from mmcv.runner import BaseModule, Sequential
+
+from mmocr.models.builder import HEADS
+from .head_mixin import HeadMixin
+
+
+@HEADS.register_module()
+class EASTHead(HeadMixin, BaseModule):
+
+    def __init__(self,
```
xinke-wang, line 12: No docstring at current version, may need to be added later.

xinke-wang:
Unittest needs to be added later.

mmocr/models/textdet/dense_heads/east_head.py
Mountchicken: You are right, This is indeed weird.

Mountchicken:
Still working on the loss non-drop problem. When it can be reproduced, I'll solve problems above.

PR #962, opened 2022-04-21
[Enhancement] Support dynamic threshold range in eval_hmean

gaotongxiao:
## Motivation

The current implementation of `eval_hmean` only provides a parameter to set the minimum score with which the boundaries are used for evaluation. However, it always evaluates the filtered boundaries on a fixed set of threshold values `[0.3, 0,4, ..., 0.9]` no matter what the minimum score is, which is an inflexible and illogical design. There is also no entry in the config to customize such a behavior.

## Modification

This PR replaces `score_thr` with `min_score_thr, max_score_thr, step` so that users can configure the search space via input parameters. `IcdarDataset.evaluate()` is modified to allow these three parameters to be customized through the config file. 

For example, by adapting the following snippet in the config, one can evaluate the model's output on a list of boundary score thresholds `[0.1, 0.2, 0.3, 0.4, 0.5]` and get the best score from them.

```python
evaluation = dict(
    interval=100,
    metric='hmean-iou',
    min_score_thr=0.1,
    max_score_thr=0.5,
    step=0.1)
```

## BC-breaking (Optional)

None

### DIFF
--- a/configs/textdet/dbnet/dbnet_r18_fpnc_1200e_icdar2015.py
+++ b/configs/textdet/dbnet/dbnet_r18_fpnc_1200e_icdar2015.py
@@ -30,4 +30,9 @@ data = dict(
         datasets=test_list,
         pipeline=test_pipeline_1333_736))
 
-evaluation = dict(interval=100, metric='hmean-iou')
+evaluation = dict(
+    interval=100,
+    metric='hmean-iou',
+    min_score_thr=0.1,
+    max_score_thr=0.5,
+    step=0.1)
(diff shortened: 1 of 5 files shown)


Mountchicken (commented):
- `eval_hmean` is also used in [TextDetDataset](https://github.com/open-mmlab/mmocr/blob/main/mmocr/datasets/text_det_dataset.py#L113). We'd better modify it too.
- BTW, One thing that confuses me is that I didn't see these score_thr parameters that can be passed to eval_hmean in the config files, do users have to modify the source code to achieve this?

gaotongxiao:
1. Sounds good
2. In fact, all the extra keys in the `evaluation` field of config are passed to `dataset.evaluate`. We just need to make sure `dataset.evaluate` parses the parameter correctly.

Mountchicken (approved):

Issue #965, opened 2022-04-23
ModuleNotFoundError: No module named 'mmdet.datasets.pipelines.formatting'


Hidayathamir:
**Checklist**

1. Yes. I have searched related issues but cannot get the expected help.
2. Yes. The bug has not been fixed in the latest version.

**Describe the bug**
I have not done anything yet, just clone the repo, build docker image, run the image, then do text detection that [documentation](https://mmocr.readthedocs.io/en/latest/demo.html#example-1-text-detection) tell me, but I got this error `ModuleNotFoundError: No module named 'mmdet.datasets.pipelines.formatting'`

**Reproduction**

1. What command or script did you run?
```
git clone https://github.com/open-mmlab/mmocr.git

cd mmocr/

docker build -t mmocr docker/

docker run -itd --gpus all --name mmocr --shm-size=8g -v /datadrive/mmocr:/mmocr mmocr
```
I am attach to running container with vscode extention [Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers). Inside the container, I run this command.
```none
python mmocr/utils/ocr.py demo/demo_text_det.jpg --output demo/det_out.jpg --det TextSnake --recog None --export demo/
```
Then I got this error `ModuleNotFoundError: No module named 'mmdet.datasets.pipelines.formatting'`.

2. Did you make any modifications on the code or config? No.
3. What dataset did you use? `demo/demo_text_det.jpg`.

**Environment**

1. Please run `python mmocr/utils/collect_env.py` to collect necessary environment information and paste it here.
```
root@43f06e5422f8:/mmocr# python mmocr/utils/collect_env.py
sys.platform: linux
Python: 3.7.7 (default, May  7 2020, 21:25:33) [GCC 7.3.0]
CUDA available: True
GPU 0: Tesla T4
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 10.1, V10.1.24
GCC: gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0
PyTorch: 1.6.0
PyTorch compiling details: PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
  - Intel(R) Math Kernel Library Version 2020.0.1 Product Build 20200208 for Intel(R) 64 architecture applications
[...]

PR #966, opened 2022-04-23
[Update] Update the version requirement of mmdet in docker

Mountchicken:
#965


xinke-wang:
I think we may also need to update the dependencies [here](https://mmocr.readthedocs.io/en/latest/install.html). The main branch requires a mmdet version >= 2.21. Also, there is a typo here, the `master` should be revised to `main` for our repo.

xinke-wang (approved):

docs/en/install.md
```diff
@@ -16,8 +16,8 @@ MMOCR has different version requirements on MMCV and MMDetection at each release
 
 | MMOCR        | MMCV                   | MMDetection               |
 | ------------ | ---------------------- | ------------------------- |
-| master       | 1.3.8 <= mmcv <= 1.6.0 | 2.14.0 <= mmdet <= 3.0.0  |
-| 0.5.0        | 1.3.8 <= mmcv <= 1.5.0 | 2.14.0 <= mmdet <= 3.0.0  |
+|   main       | 1.3.8 <= mmcv <= 1.6.0 | 2.21.0 <= mmdet <= 3.0.0  |
+| 0.5.0        | 1.3.8 <= mmcv <= 1.5.0 | 2.21.0 <= mmdet <= 3.0.0  |
```
gaotongxiao, line 20: #944 was merged after the release of 0.5.0. 0.5.0 is still compatible with mmdet>=2.14.0

gaotongxiao (changes requested):

docs/en/install.md
gaotongxiao: And there are extra spaces before "main"

gaotongxiao (commented):
Forgot to mention a place needed to be changed as well: https://github.com/open-mmlab/mmocr/blob/main/mmocr/__init__.py#L59

Mountchicken:
> Forgot to mention a place needed to be changed as well: https://github.com/open-mmlab/mmocr/blob/main/mmocr/__init__.py#L59

Done

gaotongxiao (approved):

PR #981, opened 2022-04-27
[Feature] Support computing mean scores in UniformConcatDataset

gaotongxiao:
## Motivation

Since text recognition models are usually evaluated on multiple datasets, it's hard to compare the model's performance across epochs without a unified indicator such as the mean scores. This PR supports `get_mean` in `UniformConcatDataset`. When it's on, the mean score of `{metric_name}` across concatenated datasets will be added to the evaluation results with the name `mean_{metric_name}`.

## Modification

Modified `mmdet.datasets.ConcatDataset.evaluate()` to compute the mean scores when both `self.separate_eval` and `self.get_mean` are True. Also disabled evaluating the datasets as a whole as this feature was not verified on MMOCR. We don't want to make excessive changes in this PR.

## BC-breaking (Optional)

No

## Todo

- [ ] Rename `get_mean` if we have a better suggestion
- [ ] Enable `get_mean` in all recog configs

### DIFF
--- a/mmocr/datasets/uniform_concat_dataset.py
+++ b/mmocr/datasets/uniform_concat_dataset.py
@@ -1,6 +1,9 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import copy
+from collections import defaultdict
 
+import numpy as np
+from mmcv.utils import print_log
 from mmdet.datasets import DATASETS, ConcatDataset, build_dataset
 
 from mmocr.utils import is_2dlist, is_type_list
@@ -30,6 +33,7 @@ class UniformConcatDataset(ConcatDataset):
     def __init__(self,
                  datasets,
                  separate_eval=True,
+                 get_mean=False,
                  pipeline=None,
                  force_apply=False,
                  **kwargs):
@@ -56,6 +60,12 @@ class UniformConcatDataset(ConcatDataset):
                 new_datasets = datasets
         datasets = [build_dataset(c, kwargs) for c in new_datasets]
         super().__init__(datasets, separate_eval)
+        self.get_mean = get_mean
+        if get_mean:
+            if len(set([type(ds) for ds in self.datasets])) != 1:
+                raise NotImplementedError(
+                    'To compute mean evaluation scores, all datasets'
+                    'must have the same type')
 
     @staticmethod
     def _apply_pipeline(datasets, pipeline, force_apply=False):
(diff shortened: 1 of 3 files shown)


mmocr/datasets/uniform_concat_dataset.py
```diff
@@ -30,6 +33,7 @@ class UniformConcatDataset(ConcatDataset):
     def __init__(self,
                  datasets,
                  separate_eval=True,
+                 get_mean=False,
```
xinke-wang, line 36: No docstring for `get_mean`

gaotongxiao:
Any other comments? If not, I'll start to update the configs.

xinke-wang:
> Any other comments? If not, I'll start to update the configs.

Wait a moment. I am still testing this code.

xinke-wang:
I have some minor concerns about the current evaluation process:

- Current log of recognition is not clear especially when the number of test datasets increases, for example, the index `0_`, `1_`, `2_` ... makes it difficult to find the specific subset. (Not sure if it is necessary to fix in this PR)

screenshot-20220428-121009

In comparison, the det log prints details of each dataset in a clearer way. So the users can check the performance of a specific dataset easily.

image

- Using different dataset type in detection task cannot trigger the `NotImplementedError`, instead, it raises `AttributeError: UniformConcatDataset: 'IcdarDataset' object has no attribute 'flag'`.

Using the following toy data config to reproduce the issue
```python
root = 'tests/data/toy_dataset'

# dataset with type='TextDetDataset'
train1 = dict(
    type='TextDetDataset',
    img_prefix=f'{root}/imgs',
    ann_file=f'{root}/instances_test.txt',
    loader=dict(
        type='HardDiskLoader',
        repeat=4,
        parser=dict(
            type='LineJsonParser',
            keys=['file_name', 'height', 'width', 'annotations'])),
    pipeline=None,
    test_mode=False)

# dataset with type='IcdarDataset'
train2 = dict(
    type='IcdarDataset',
    ann_file=f'{root}/instances_test.json',
    img_prefix=f'{root}/imgs',
    pipeline=None)

test = dict(
    type='TextDetDataset',
    img_prefix=f'{root}/imgs',
    ann_file=f'{root}/instances_test.txt',
    loader=dict(
        type='HardDiskLoader',
        repeat=1,
        parser=dict(
            type='LineJsonParser',
            keys=['file_name', 'height', 'width', 'annotations'])),
    pipeline=None,
    test_mode=True)

train_list = [train1, train2]

test_list = [test, train2]

```

gaotongxiao:
@xinke-wang Unfortunately, even though we can print the results for each dataset, there is no way to get rid of the summary at the end of the evaluation, which can lead to duplicate outputs. To streamline the information, I can make another PR after which users can choose the evaluation metric(s) to report by customizing the config.

xinke-wang (approved):

PR #983, opened 2022-04-27
[Enhancement] Add SynthText dataset config

xinke-wang:
## Motivation

- Pre-training text detectors on synthetic datasets is a widely-used strategy for improving performance; this PR adds SynthText dataset configs.
- Currently, there are two runtime settings which are `configs/_base_/default_runtime.py` and `configs/_base_/runtime_10e.py`. The only difference between these configs is the checkpoint `interval` param. However, this param should be relative to the total epoch/iteration number. In addition, the default setting `interval=10` saves too many checkpoints for some tasks with a long schedule, such as `1200 epochs` or `100,000 iterations`. Therefore, it might be more reasonable to move this param to schedule configs.

## Modification

- Add SynthText dataset config in `configs/_base_/det_datasets/synthtext.py`.
- Move the `checkpoint_config` to schedule config files, and set a more reasonable default `interval` depending on the total epoch/iteration.
- Delete the duplicated `runtime_10e.py` and modify all configs that use `runtime_10e.py` to `default_runtime.py`.

## BC-breaking (Optional)

One thing that should be noted is that PR #949 replaces the `imgaug.Keypoint` with `imgaug.Polygon` for polygon data annotations, and the `ImgAug` will no longer support `invalid` polygons. However, there are a considerable number of invalid polygons in the SynthText annotations. Therefore, users should add a `make_valid` function/transform to convert these invalid polygons by eliminating self-crossing or self-touching parts; otherwise, these samples will be abandoned.

## Use cases (Optional)

For **reviewer** only:

To use the SynthText data on sproject, here are some notes:
1. There are only images stored on ceph, so you have to download the annotations to the local driver from [here](https://mmocr.readthedocs.io/en/latest/datasets/det.html#synthtext), or I have already downloaded one which is stored in `share_data/.../synthtext/instances_training.lmdb`. Then you have to set the `file_client_args=dict(backend='petrel')` in training pipelilnes.
2. The img path on ceph is a bit different from our doc, hence you have to modify the `synthtext.py` as follows:
```python
dataset_type = 'TextDetDataset'
data_root = 'openmmlab:s3://openmmlab/datasets/ocr/recog'

train = dict(
    type=dataset_type,
    ann_file='data/synthtext/instances_training.lmdb',
    loader=dict(
        type='AnnFileLoader',
        repeat=1,
        file_format='lmdb',
        parser=dict(
            type='LineJsonParser',
            keys=['file_name', 'height', 'width', 'annotations'])),
    img_prefix=f'{data_root}/SynthText/',
    pipeline=None)

train_list = [train]
test_list = [train]

```


configs/textdet/dbnet/dbnet_r50dcnv2_fpnc_100k_iters_synthtext.py
```diff
@@ -0,0 +1,35 @@
...
+    '../../_base_/det_pipelines/dbnet_pipeline.py'
+]
+
+train_list = {{_base_.train_list}}
+test_list = {{_base_.test_list}}
+
+train_pipeline_r50dcnv2 = {{_base_.train_pipeline_r50dcnv2}}
+test_pipeline_4068_1024 = {{_base_.test_pipeline_4068_1024}}
+
+load_from = 'checkpoints/textdet/dbnet/res50dcnv2_synthtext.pth'
```
gaotongxiao, line 15: no `load_from` here

configs/textdet/dbnet/dbnet_r50dcnv2_fpnc_100k_iters_synthtext.py
```diff
@@ -0,0 +1,35 @@
...
+
+data = dict(
+    samples_per_gpu=8,
+    workers_per_gpu=4,
```
gaotongxiao, line 19:
```suggestion
    samples_per_gpu=16,
    workers_per_gpu=8,
```
Suggested setting (not necessarily true for synthtext)

configs/textdet/dbnet/dbnet_r18_fpnc_100k_iters_synthtext.py
```diff
@@ -0,0 +1,33 @@
+_base_ = [
+    '../../_base_/default_runtime.py',
+    '../../_base_/schedules/schedule_sgd_100k_iters.py',
+    '../../_base_/det_models/dbnet_r18_fpnc.py',
```
gaotongxiao, line 4: Shall we create a config for r18dcn?

configs/textdet/dbnet/dbnet_r50dcnv2_fpnc_100k_iters_synthtext.py
```diff
@@ -0,0 +1,35 @@
...
+    val=dict(
+        type='UniformConcatDataset',
+        datasets=test_list,
+        pipeline=test_pipeline_4068_1024),
+    test=dict(
+        type='UniformConcatDataset',
+        datasets=test_list,
+        pipeline=test_pipeline_4068_1024))
+
+evaluation = dict(interval=999999, metric='hmean-iou')  # do not evaluate
```
gaotongxiao, line 35: @zhouzaida any better suggestion on disabling evaluation in training?

configs/textdet/dbnet/dbnet_r18_fpnc_100k_iters_synthtext.py
xinke-wang: > Shall we create a config for r18dcn?

Yes, I think so, since the official code use dcn. Shall we directly override the r18 with r18dcn or create another one?

Mountchicken (approved):

PR #984, opened 2022-04-27
[Fix] Update installation step of `albumentations`

gaotongxiao:
## Motivation

By default, installing `albumentations` by pip has dependency on `opencv-python-headless` even if `opencv-python` has already been installed. This could cause different opencv versions installed in the environment and lead to unexpected errors. `albumentations` has updated the installation step, which changes `pip install albumentations --no-binary imgaug,albumentations` to `pip install -U albumentations --no-binary qudida,albumentations`. See its [docs](https://albumentations.ai/docs/getting_started/installation/#note-on-opencv-dependencies) for details.

If simply use `pip install albumentations>=1.1.0 --no-binary qudida,albumentations` may cause some error or installed `opencv-python-headless` simultaneously. so it is suggested to use `pip install -r requirements/albu.txt` or `pip install -U albumentations --no-binary qudida,albumentations` to install albu.

## Reference
https://github.com/open-mmlab/mmdetection/pull/7143
https://github.com/open-mmlab/mmdetection/pull/5868

### DIFF
--- a/docs/en/install.md
+++ b/docs/en/install.md
@@ -16,7 +16,7 @@ MMOCR has different version requirements on MMCV and MMDetection at each release
 
 | MMOCR        | MMCV                   | MMDetection               |
 | ------------ | ---------------------- | ------------------------- |
-|   main       | 1.3.8 <= mmcv <= 1.6.0 | 2.21.0 <= mmdet <= 3.0.0  |
+| main         | 1.3.8 <= mmcv <= 1.6.0 | 2.21.0 <= mmdet <= 3.0.0  |
 | 0.5.0        | 1.3.8 <= mmcv <= 1.5.0 | 2.14.0 <= mmdet <= 3.0.0  |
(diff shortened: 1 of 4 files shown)


Mountchicken (approved):
Working on Colab now. One thing I don't understand is why not just change the content in optional.txt to `albumentations>=1.1.0 --no-binary qudida, albumentations`

gaotongxiao:
@Mountchicken The packages in `optional.txt` will be installed when the user runs `pip install -r requirements.txt`. However, pip does not guarantee the installation order of packages requested in the same command, so `albumentations` could be installed before `opencv-python`. Then according to [docs](https://albumentations.ai/docs/getting_started/installation/#note-on-opencv-dependencies), `albumentations` still installs `opencv-python-headless` if no OpenCV is found in the environment. We now intentionally put `albumentations` to the end of the installation to prevent it from happening.

PR #989, opened 2022-04-29
[Enhancement] Allow choosing metrics to report in text recognition tasks

gaotongxiao:
## Motivation

Though MMOCR provides 6 evaluation metrics for text recognition tasks, there was no way to configure the metric(s) to evaluate the model, which usually results in an unnecessarily ong performance report. This PR enables the configuration entrance and users now can evaluate models on a subset of metrics selectively.

## Use Cases

Take the evaluation config in `configs/textrecog/crnn/crnn_academic_dataset.py` as an example. 

```python
evaluation = dict(interval=1, metric='acc')
```

The recognition metric `acc` is short for a collection of 6 metrics. One can expect a full performance report at the end of an evaluation cycle:

```python
{'0_char_recall': 0.0484, '0_char_precision': 0.6, '0_word_acc': 0.0, '0_word_acc_ignore_case': 0.0, '0_word_acc_ignore_case_symbol': 0.0, '0_1-N.E.D': 0.0525}
```

After this PR, one can specify a metric or a group of metrics to report in a way like:

```python
evaluation = dict(interval=1, metric=['word_acc_ignore_case', 'one_minus_ned'])
```

The result looks like:

```python
{'0_word_acc_ignore_case': 0.0, '0_1-N.E.D': 0.0525}
```

In testing, one can also specify the metrics to evaluate in the command line:

```bash
python tools/test.py configs/textrecog/crnn/crnn_toy_dataset.py crnn.pth --eval word_acc_ignore_case one_minus_ned
```

## BC-breaking:

None. We still hold the compatibility with `acc` metric for now.

### DIFF
--- a/mmocr/core/evaluation/ocr_metric.py
+++ b/mmocr/core/evaluation/ocr_metric.py
@@ -4,6 +4,8 @@ from difflib import SequenceMatcher
 
 from rapidfuzz import string_metric
 
+from mmocr.utils import is_type_list
+
 
 def cal_true_positive_char(pred, gt):
     """Calculate correct character number in prediction.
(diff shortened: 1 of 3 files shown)


Mountchicken:
A strange thing happens. When I use `evaluation = dict(interval=1, metric='acc', show_mean_scores=True)`, there is no output of the mean value. However, if I set `show_mean_scores=True` as default in UniforConcatDataset, then I can get the mean value. It seems that `show_mean_scores` is not passed into the UniforConcatDataset.

Mountchicken (approved):
Tested on Colab, and everything works fine.

PR #998, opened 2022-05-05
[Fix] Fix lint_only in CircleCI

gaotongxiao:
According to the current config, full tests are triggered when CircleCI does not detect the changes in a set of paths. However, it was logically wrong since if a commit changes both the docs and code, `lint_only` will still be true and no full tests would be performed. This PR flips the mapping around to ensure no undesired skips when the case above happens.


gaotongxiao:
@lvhan028 Your configuration might also need to be updated

xinke-wang (approved):

PR #1006, opened 2022-05-09
[Feature] Add ArT

xinke-wang:
## Motivation

ArT is a widely-used arbitrary scene text dataset that already contains Total-text and Ctw-1500, and extra curved scene text. Also, it refines the annotations of CTW-1500, and all Chinese texts are annotated in line-level while English samples are word-level.

## Modification

add `art_converter.py` for textdet and textrecog

## Use cases (Optional)

See doc


docs/en/datasets/recog.md
```diff
@@ -1116,3 +1117,40 @@ should be as follows:
...
+- Step2: Generate `train_label.jsonl` and `val_label.jsonl` (optional). Since the test annotations are not publicly available, you may specify `--val-ratio` to split the dataset. E.g., if val-ratio is 0.2, then 20% of the data are left out as the validation set in this example.
+
+  ```bash
+  # Annotations of ArT test split is not publicly available, split a validation set by adding --val-ratio 0.2
+  python tools/data/textrecog/art_converter.py PATH/TO/art
+  ```
+
+- After running the above codes, the directory structure should be as follows:
+
+  ```te
```
Mountchicken, line 1151: text

PR #1018, opened 2022-05-17
[Fix] inplace operator "+=" will cause RuntimeError when model backward

garvan2021:
ref:https://discuss.pytorch.org/t/runtimeerror-one-of-the-variables-needed-for-gradient-computation-has-been-modified-by-an-inplace-operation-torch-cuda-floattensor-1-1-256-256-which-is-output-0-of-relubackward0-is-at-version-1-expected-version-0-instead/143490


Mountchicken:
Hi @garvan2021 
Thanks for pointing it out. I noticed a similar operation in `fpn_cat` which may also cause this problem. Please check if this should be fixed too.
- [ ] [fpn_cat](https://github.com/open-mmlab/mmocr/blob/main/mmocr/models/textdet/necks/fpn_cat.py#L153)

garvan2021:
weird, its ok to use inplace operation when the neck is `fpn_cat`

I changed the [fpnf](https://github.com/open-mmlab/mmocr/blob/3059c97dc53b000a1509916ad30223329d926b02/configs/_base_/det_models/psenet_r50_fpnf.py#L39) as follow:
>model_quad = dict(
    type='PSENet',
    backbone=dict(
        type='mmdet.ResNet',
        depth=50,
        num_stages=4,
        out_indices=(0, 1, 2, 3),
        frozen_stages=-1,
        norm_cfg=dict(type='SyncBN', requires_grad=True),
        init_cfg=dict(type='Pretrained', checkpoint='torchvision://resnet50'),
        norm_eval=True,
        style='caffe'),
    neck=dict(
        type='FPNC', in_channels=[256, 512, 1024, 2048], lateral_channels=256),
    bbox_head=dict(
        type='PSEHead',
        in_channels=[256],
        out_channels=7,
        loss=dict(type='PSELoss'),
        postprocessor=dict(type='PSEPostprocessor', text_repr_type='quad')),
    train_cfg=None,
    test_cfg=None)

here is my training command:
>python tools/train.py configs/textdet/psenet/psenet_r50_fpnf_600e_icdar2015.py --work-dir psenet

Mountchicken:
Strange indeed. But I think it should still be modified to prevent potential inplace bugs.

Mountchicken (approved):

gaotongxiao (approved):

PR #1060, opened 2022-06-08
simplify normalized edit distance calculation

maxbachmann:
## Motivation

Replace usage of deprecated APIs and simplify implementation.

## Modification

The string_metric module in rapidfuzz is deprecated since v2.0.0 in addition the new module `rapidfuzz.distance.Levenshtein` provides a function to directly calculate the normalized Levenshtein distance. This can be used to simplify the implementation in mmocr.

## Checklist

**Before PR**:

**After PR**:

- [x] If the modification has potential influence on downstream or other related projects, this PR should be tested with some of those projects.
- [x] CLA has been signed and all committers have signed the CLA in this PR.

### DIFF
--- a/mmocr/core/evaluation/ocr_metric.py
+++ b/mmocr/core/evaluation/ocr_metric.py
@@ -2,7 +2,7 @@
 import re
 from difflib import SequenceMatcher
 
-from rapidfuzz import string_metric
+from rapidfuzz.distance import Levenshtein
 
 from mmocr.utils import is_type_list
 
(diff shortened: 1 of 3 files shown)


maxbachmann:
In addition it would be possible to replace:
```python
for opt, _, _, s2, e2 in all_opt.get_opcodes():
```
with
```python
for opt, _, _, s2, e2 in Levenshtein.opcodes():
```
in case you prefer to calculate the opcodes based on the Levenshtein distance instead of the Ratcliff and Obershelp algorithm.

gaotongxiao (commented):
Thanks for your contribution! As this change relies on a newer API, it would be better to update requirements/runtime.txt as well to restrict `rapidfuzz`'s minimum version to `2.0.0`.

maxbachmann:
> Thanks for your contribution! As this change relies on a newer API, it would be better to update requirements/runtime.txt as well to restrict rapidfuzz's minimum version to 2.0.0.

Updated accordingly

gaotongxiao (approved):

Issue #1143, opened 2022-07-09
bounding box sort clockwise function is incorrect


wybryan:
**Describe the bug**

from mmocr.utils.box_util import sort_points
poly = [[1668,1748],[1704,1749],[1742,1898],[1705,1898]]
sorted_pts = order_points(np.array(poly))
print(sorted_pts)

the MMOCR function sort_points claims to sort arbitory points in clockwise order. 

BUT the above code produce sorted points as 
[1704, 1749]
[1668, 1748]
[1705, 1898]
[1742, 1898]

which is completely incorrect. The expected order should be:

[1668,1748]
[1704,1749]
[1742,1898]
[1705,1898]

**Reproduction**

1. What command or script did you run?

from mmocr.utils.box_util import sort_points
poly = [[1668,1748],[1704,1749],[1742,1898],[1705,1898]]
sorted_pts = order_points(np.array(poly))
print(sorted_pts)

The MMOCR version used is 0.6.0.

**Environment**

1. Please run `python mmocr/utils/collect_env.py` to collect necessary environment information and paste it here.

sys.platform: linux
Python: 3.9.12 (main, Apr  5 2022, 06:56:58) [GCC 7.5.0]
CUDA available: False
GCC: gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
PyTorch: 1.9.0+cu111
PyTorch compiling details: PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
  - Intel(R) oneAPI Math Kernel Library Version 2021.4-Product Build 20210904 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v2.1.2 (Git Hash 98be7e8afa711dc9b66c8ff3504129cb82013cdb)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - NNPACK is enabled
  - CPU capability usage: AVX2
[...]

PR #1146, opened 2022-07-11
[Bug fix] box points ordering

wybryan:

## Motivation

This PR tries to fix bounding box point ordering issue I reported at https://github.com/open-mmlab/mmocr/issues/1143

## Modification

bug fix

## BC-breaking (Optional)

Don't think so.

### DIFF
--- a/configs/textdet/psenet/metafile.yml
+++ b/configs/textdet/psenet/metafile.yml
@@ -38,7 +38,7 @@ Models:
           hmean-iou: 0.807
     Weights: https://download.openmmlab.com/mmocr/textdet/psenet/psenet_r50_fpnf_600e_icdar2015-c6131f0d.pth
 
-  - Name: psenet_r50_fpnf_600e_icdar2015_with_pretrain
+  - Name: psenet_r50_fpnf_600e_icdar2015
     In Collection: PSENet
     Config: configs/textdet/psenet/psenet_r50_fpnf_600e_icdar2015.py
     Metadata:
--- a/docs/en/install.md
+++ b/docs/en/install.md
@@ -97,7 +97,7 @@ See [here](https://pypi.org/project/mmdet/) for different versions of `mmdet`.
 pip install mmdet
 ```
 
-Optionally you can choose to install `mmdet` following the official [installation guide](https://github.com/open-mmlab/mmdetection/blob/master/docs/get_started.md).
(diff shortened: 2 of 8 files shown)


gaotongxiao:
Thanks for your PR! However, the tests did not seem to pass.

In fact, we have long been aware of the bug of `sort_points` and the existence of its alternative in `mmocr.datasets.pipelines.box_utils.sort_vertex`, but did not have time to compare their implementation and eliminate the duplicate. Could you try `sort_vertex` as well see if it can solve your issue?

wybryan:
> Thanks for your PR! However, the tests did not seem to pass.
> 
> In fact, we have long been aware of the bug of `sort_points` and the existence of its alternative in `mmocr.datasets.pipelines.box_utils.sort_vertex`, but did not have time to compare their implementation and eliminate the duplicate. Could you try `sort_vertex` as well see if it can solve your issue?



> Thanks for your PR! However, the tests did not seem to pass.
> 
> In fact, we have long been aware of the bug of `sort_points` and the existence of its alternative in `mmocr.datasets.pipelines.box_utils.sort_vertex`, but did not have time to compare their implementation and eliminate the duplicate. Could you try `sort_vertex` as well see if it can solve your issue?

I'll try sort_vertext.

regarding the test failure, I notice there is another error in the test itself: look at code 

points = np.array([[1, 1], [1, -1], [-1, 1], [-1, -1]])
target = np.array([[-1, -1], [-1, 1], [1, 1], [1, -1]])
assert np.allclose(target, sort_points(points))

Isn't the target is wrong? it is anti-clockwise, not clock-wise. This is the reason the proposed solution 'fail' the test, where the test reference is not right...

wybryan:
> Hi, the test is now passing when incorrect test reference gets fixed now.

Do you know the test scripts for sort_vertex so I can test it also?

It is my impression that sort_vertex implementation seems quite cumbersome even if it is correct, therefore I'd like to use same test suite to test my proposal so see if it all works, I see there is opportunity to optimize such point ordering function for mmocr once for all.

gaotongxiao:
@wybryan The target might or might not be wrong depending on the coordinate system of the external library that uses the polygon. We need a quick investigation before we set up a gold standard for all the relevant implementations. And `sort_vertex` doesn't come with a test suite.

PR #1282, opened 2022-08-15
[Inferencer] New inference API

gaotongxiao:
Based on #1237

**Note**: `BaseInferencer` is intended to be used in MMEngine. `BaseMMOCRInferencer` is the base class for all inferencers in MMOCR.

This PR serves as a demo of the design of Inferencer. With the help of `BaseInferencer`, `TextDetInferencer` and `TextRecInferencer` share the same interface with only minimal changes. As shown below, both inferencers support nearly all the options in the original `ocr.py` except for the batch inference, which will be implemented if time permitted.

```python
"""
    Args:
        model (str or ConfigType): Model config or the path to it.
        ckpt (str, optional): Path to the checkpoint.
        device (str, optional): Device to run inference. If None, the best
            device will be automatically used.
        show (bool): Whether to display the image in a popup window.
            Defaults to False.
        wait_time (float): The interval of show (s). Defaults to 0.
        draw_pred (bool): Whether to draw predicted bounding boxes.
            Defaults to True.
        pred_score_thr (float): Minimum score of bboxes to draw.
            Defaults to 0.3.
        img_out_dir (str): Output directory of images. Defaults to ''.
        pred_out_file: File to save the inference results. If left as empty, no
            file will be saved.
        print_result (bool): Whether to print the result.
            Defaults to False.
"""
```

Note that this design allows model-independent parameters (all the parameters above except for `model`, `ckpt` and `device`) to be specified in `__init__` (at initialization) and/or `__call__` (at runtime). Parameters specified during initialization will be regarded as defaults, which can be temporarily overwritten by the runtime specification.

## Usage

```python
from mmocr.apis.inferencers import TextRecInferencer
from mmocr.utils import register_all_modules
register_all_modules()

inferencer = TextRecInferencer('configs/textrecog/satrn/satrn_small.py', 'satrn_small.pth', show=True)

# The visualized recognition result will be saved to "inference_out/", and the result will not be displayed
inferencer('demo/demo_text_recog.jpg', img_out_dir='inference_out', show=False)

# The visualized recognition result will be saved to "inference_out/", and the result will not be displayed
inferencer(['demo/demo_text_recog.jpg', 'demo/demo_text_det.jpg'])

# Or its equivalent form
import mmcv
img1 = mmcv.imread('demo/demo_text_det.jpg')
img2 = mmcv.imread('demo/demo_text_recog.jpg')
inferencer([img1, img2])
```

## TODO

- [ ] More docstring and type hints
- [x] `KIEInferencer`
- [x] `MMOCRInferencer` which acts exactly like `ocr.py`
- [ ] batch inference
- [ ] CLI interface
- [ ] Unit tests (after this design has been accepted)


mmocr/apis/inferencers/base_inferencer.py
```diff
@@ -0,0 +1,307 @@
...
+
+    def forward(self, inputs: InputsType) -> PredType:
+        with torch.no_grad():
+            return self.model.test_step(inputs)
+
+    def visualize(self,
+                  inputs: InputsType,
+                  preds: PredType,
+                  show: bool = False,
+                  wait_time: int = 0,
```
Mountchicken, line 207:
```suggestion
                  wait_time: float = 0.,
```

PR #1296, opened 2022-08-20
[Config] Rename DB, DB++ and DRRG

gaotongxiao:


xinke-wang (approved):

Harold-lkk (approved):

ZwwWayne:
Suggest unifying the convention of resnets, we can simply use r50, r18 in all the codebases if they are the conventional resnets.

PR #1338, opened 2022-08-29
[Fix] Fix LMDB dataset

xinke-wang:
This PR fixes：

1. `super().__init__()` will concat `self.ann_file` and `self.data_root`, however, the first `_get_env()` will be called before `super().__init__()`, hence, the `ann_file` cannot be found if it is not an absolute path. And this issue cannot be solved by simply move the `super().__init__()` to the beginning of `init()`, because the `super().__init()__` relies on `self.total_number` which is obtained by `_get_env()`. Therefore, this PR adds a `root` arg to help the `_get_env()` to get the correct path in `init()`.

### DIFF
--- a/mmocr/datasets/recog_lmdb_dataset.py
+++ b/mmocr/datasets/recog_lmdb_dataset.py
@@ -80,7 +80,7 @@ class RecogLMDBDataset(BaseDataset):
         self.parser = TASK_UTILS.build(parser_cfg)
         self.ann_file = ann_file
         self.deprecated_format = False
-        env = self._get_env()
+        env = self._get_env(root=data_root)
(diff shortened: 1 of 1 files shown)


mmocr/datasets/recog_lmdb_dataset.py
```diff
@@ -181,7 +181,7 @@ def _get_env(self):
             raise ImportError(
                 'Please install lmdb to enable RecogLMDBDataset.')
         return lmdb.open(
-            self.ann_file,
+            osp.join(root, self.ann_file),
```
Harold-lkk, line 184: judge ann file whether abs path

Harold-lkk (approved):

PR #1430, opened 2022-10-08
[main] Upgrade pre commit hooks

Harold-lkk:


tools/data/common/labelme_converter.py
```diff
@@ -205,7 +205,7 @@ def parse_args():
         choices=['jsonl', 'txt'])
     parser.add_argument(
         '--warp',
-        help='Store warpped img for recognition task',
+        help='Store wrapped img for recognition task',
```
gaotongxiao, line 208: it should be "warped"

gaotongxiao (approved):

Issue #1426, opened 2022-10-05
dev-1.x版本，用dbnet训练icdar2017mlt的问题


ConflictPeople:
您好，我想问下关于dev-1.x版本训练数据集的问题。
我将     dict(type='FixInvalidPolygon', mode='fix', min_poly_points=3)   添加到   train_pipeline  中，训练出现错误：
           Exception: Cannot determine whether the polygon is inside the image, because it contains no points.
不知道是什么问题，数据集转换是用官方提供的迁移代码，应该是没有问题的。
具体的错误如下：
10/05 21:19:27 - mmengine - INFO - Epoch(train) [1][260/1200]  lr: 7.0000e-03  eta: 2 days, 7:02:27  time: 0.8387  data_time: 0.1538  memory: 3608  loss_prob: 2.1957  loss_thr: 0.8966  loss_db: 0.3960  loss: 3.4883
10/05 21:19:32 - mmengine - INFO - Epoch(train) [1][265/1200]  lr: 7.0000e-03  eta: 2 days, 7:02:27  time: 0.8552  data_time: 0.1697  memory: 3608  loss_prob: 2.0312  loss_thr: 0.8293  loss_db: 0.3572  loss: 3.2177
10/05 21:19:35 - mmengine - INFO - Epoch(train) [1][270/1200]  lr: 7.0000e-03  eta: 2 days, 6:54:57  time: 0.7768  data_time: 0.0946  memory: 3608  loss_prob: 2.1056  loss_thr: 0.8052  loss_db: 0.3712  loss: 3.2820
10/05 21:19:39 - mmengine - INFO - Epoch(train) [1][275/1200]  lr: 7.0000e-03  eta: 2 days, 6:54:57  time: 0.7252  data_time: 0.0585  memory: 3608  loss_prob: 2.0963  loss_thr: 0.8200  loss_db: 0.3674  loss: 3.2836
10/05 21:19:43 - mmengine - INFO - Epoch(train) [1][280/1200]  lr: 7.0000e-03  eta: 2 days, 6:52:15  time: 0.8068  data_time: 0.1522  memory: 3608  loss_prob: 2.1957  loss_thr: 0.8609  loss_db: 0.3913  loss: 3.4479
[...]

PR #1434, opened 2022-10-09
[Fix] Skip invalud augmented polygons in ImgAugWrapper

gaotongxiao:
Fixes https://github.com/open-mmlab/mmocr/issues/1426

# Motivation

Sometimes ImgAug could produce some invalid polygons after applying augmentations to a polygon set and may interfere with the training process. This PR checks and skips every invalid polygon and can alleviate this issue.


xinke-wang:
lint failed?

xinke-wang (approved):

PR #1462, opened 2022-10-17
Convert numpy.float32 type to python built-in float type

JunYao1020:
## Motivation

When I follow the documentation to convert the psenet model to torchserve model and start the torchserve service, the predicted image will receive a 503 error. The reason is that the float32 type of numpy will throw an exception during serialization, and the confidence type in the result predicted by pse is numpy.float32

## Modification
Convert the confidence type in the PSE post-processing file from numpy.float32 to python built-in float
image


gaotongxiao (approved):
LGTM

PR #1503, opened 2022-11-02
[Feature] Support browse_dataset.py to visualize original dataset

xinke-wang:
Now browse_dataset.py supports two modes, `model config mode` and `dataset config mode`. Users can input dataset config now, and the script will automatically select the corresponding mode.

### DIFF
--- a/tools/analysis_tools/browse_dataset.py
+++ b/tools/analysis_tools/browse_dataset.py
@@ -1,6 +1,7 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import argparse
 import os.path as osp
+from typing import Tuple
 
 import mmengine
 from mmengine.config import Config, DictAction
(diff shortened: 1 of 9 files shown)


tools/analysis_tools/browse_dataset.py
```diff
@@ -1,33 +1,82 @@
...
 
 from mmocr.registry import DATASETS, VISUALIZERS
 from mmocr.utils import register_all_modules
 
 
+# TODO: Support for printing the change in key of results
 def parse_args():
     parser = argparse.ArgumentParser(description='Browse a dataset')
-    parser.add_argument('config', help='Train config file path')
+    parser.add_argument('config', help='Path to model or dataset config.')
```
Harold-lkk, line 22: introduce some limit for dataset config

tools/analysis_tools/browse_dataset.py
```diff
@@ -37,40 +86,313 @@ def parse_args():
     return args
 
 
-def main():
+def _get_adaptive_scale(img_shape: Tuple[int, int],
```
Harold-lkk, line 89: would it be better for all images to have the same height?

Harold-lkk (approved):

PR #1521, opened 2022-11-09
[Feature] Add SVT to dataset preparer

xinke-wang:


gaotongxiao (approved):
LGTM. We just need a local test before mering

xinke-wang:
> LGTM. We just need a local test before mering

Test added. By the way, it seems something is wrong with the CI. It will be stuck at `upgrade pip`.

gaotongxiao:
@xinke-wang "need test" here doesn't mean unit tests. I was marking these PRs and would find a time to run your scripts on a clean environment and see if generated data are as expected.

Harold-lkk (approved):

PR #1506, opened 2022-11-02
[Docs] Add Chinese Guidance on How to Add New Datasets to Dataset Preparer

xinke-wang:
Add Guidance on How to Add New Datasets to Dataset Preparer (Chinese Version Only), using ICDAR 2013 dataset as an example. Also, this PR adds the IC13 dataset to the dataset preparer.

PR #1526, opened 2022-11-15
[Bugs] fix crop without padding and recog metainfo delete unuse info

Harold-lkk:
Delete category in rec json
```python
    "metainfo": {
        "dataset_type": "TextRecogDataset",
        "task_name": "textrecog",
        "category": [
            {
                "id": 0,
                "name": "text"
            }
        ]
    }
```

`TextRecogCropConverter` defaults not to padding in `crop image`

### DIFF
--- a/mmocr/datasets/preparers/data_converter.py
+++ b/mmocr/datasets/preparers/data_converter.py
@@ -476,11 +476,7 @@ class TextRecogDataConverter(BaseDataConverter):
         meta = {
             'metainfo': {
                 'dataset_type': 'TextRecogDataset',
-                'task_name': 'textrecog',
-                'category': [{
-                    'id': 0,
-                    'name': 'text'
-                }]
+                'task_name': 'textrecog'
             },
             'data_list': sample
         }
@@ -517,8 +513,8 @@ class TextRecogCropConverter(TextRecogDataConverter):
                  dumper: Dict,
                  dataset_name: str,
                  nproc: int,
-                 long_edge_pad_ratio: float = 0.1,
-                 short_edge_pad_ratio: float = 0.05,
+                 long_edge_pad_ratio: float = 0.0,
+                 short_edge_pad_ratio: float = 0.0,
                  delete: List = ['annotations']):
         super().__init__(
             splits=splits,


Harold-lkk:
merge before #1506

xinke-wang:
A question here, in 0.x version, the pad ratio was set as `0.4` and `0.2` by default. I am not sure if this matters to the test performance. Though personally, I think it should be set to `0` as this PR, maybe someone should test the model to see if it works as expected.

https://github.com/open-mmlab/mmocr/blob/26bc4713d4a451ed510a67be0a4fdd9903fd9011/mmocr/datasets/pipelines/crop.py#L87-L97

Harold-lkk:
> A question here, in 0.x version, the pad ratio was set as `0.4` and `0.2` by default. I am not sure if this matters to the test performance. Though personally, I think it should be set to `0` as this PR, maybe someone should test the model to see if it works as expected.
> 
> https://github.com/open-mmlab/mmocr/blob/26bc4713d4a451ed510a67be0a4fdd9903fd9011/mmocr/datasets/pipelines/crop.py#L87-L97

I have tested. Difference image shape will affect the performance.
In the 0.x version, the `crop image` is only used for ocr.py, because of bad performance of text detection

xinke-wang:
> > A question here, in 0.x version, the pad ratio was set as `0.4` and `0.2` by default. I am not sure if this matters to the test performance. Though personally, I think it should be set to `0` as this PR, maybe someone should test the model to see if it works as expected.
> > https://github.com/open-mmlab/mmocr/blob/26bc4713d4a451ed510a67be0a4fdd9903fd9011/mmocr/datasets/pipelines/crop.py#L87-L97
> 
> I have tested. Difference image shape will affect the performance. In the 0.x version, the `crop image` is only used for ocr.py, because of bad performance of text detection

Ok. It is noteworthy that some converters such as totaltext converter also used the `crop_image` and the default setting. I am not sure if it affects totaltext's performance.

https://github.com/open-mmlab/mmocr/blob/26bc4713d4a451ed510a67be0a4fdd9903fd9011/tools/data/textrecog/totaltext_converter.py#L302

gaotongxiao (approved):

Harold-lkk:
> crop_img

Maybe it's a bug. I check other data_converter, and they are all `crop_img(image, anno['bbox'], 0, 0)`

xinke-wang (approved):

gaotongxiao (approved):

PR #1527, opened 2022-11-15
[Fix] Fix Dataset Preparer Extract

xinke-wang:
When re-running the data preparer, it might repeatedly download the annotation files.

### DIFF
--- a/mmocr/datasets/preparers/data_obtainer.py
+++ b/mmocr/datasets/preparers/data_obtainer.py
@@ -100,8 +100,8 @@ class NaiveDataObtainer:
                 to False.
         """
         if not is_archive(src_path) and not osp.exists(dst_path):
-            # Move the file to the destination folder if it is not a zip
-            shutil.move(src_path, dst_path)
+            # Copy the file to the destination folder if it is not a zip
+            shutil.copy(src_path, dst_path)
             return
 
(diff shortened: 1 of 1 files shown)


mmocr/datasets/preparers/data_obtainer.py
```diff
@@ -99,10 +99,9 @@ def extract(self,
             delete (bool, optional): Whether to delete the zip file. Defaults
                 to False.
         """
-
-        if not is_archive(src_path):
+        if not is_archive(src_path) and not osp.exists(dst_path):
             # Move the file to the destination folder if it is not a zip
-            shutil.move(src_path, dst_path)
+            shutil.copytree(src_path, dst_path)
```
Harold-lkk, line 104: it will raise error when `scr_path` is a file and `dst_path` is a directory

Harold-lkk: Maybe add a check for whether the dst_path exists for it will raise an error if dst_path has existed through `shutil.move`

gaotongxiao (commented):
How could this change avoid repeated downloading?

xinke-wang:
> How could this change avoid repeated downloading?

`move` -> `copy` so the cached file remains after converting

gaotongxiao (approved):

Harold-lkk (approved):

PR #1528, opened 2022-11-15
[Fix] Fix wildreceipt metafile

xinke-wang:


xinke-wang:
It seems the test of the visualizer occasionally failed. @Harold-lkk

Harold-lkk (approved):

PR #1552, opened 2022-11-17
[Feature] Add config generator

gaotongxiao:
## Motivation

Currently, base configs can only be generated when using data_convertor, which is not applicable if some pre-existing annotations can be directly downloaded. This PR moves this part to an individual module and polishes the design.

This PR also revised the config for IC13 to demonstrate its usage.

PR #1556, opened 2022-11-18
[Config] Support IC15_1811 

gaotongxiao:
Based on #1552


dataset_zoo/icdar2015/textrecog.py
```diff
@@ -1,3 +1,6 @@
+# This configuration prepares the ICDAR15 1811 and 2077
+# version, and uses ICDAR13 2077 version by default.
```
Harold-lkk, line 2:
```suggestion
# version, and uses ICDAR15 2077 version by default.
```

Harold-lkk (approved):

PR #1588, opened 2022-12-02
[Enhancement] Simplify mono_gather

gaotongxiao:


mmocr/datasets/preparers/data_converter.py
```diff
@@ -116,7 +116,12 @@ def add_meta(self, sample: List) -> Dict:
             Dict: A dict contains the meta information and samples.
         """
 
-    def mono_gather(self, ann_path: str, mapping: str, split: str,
+    def mono_gather(self,
+                    ann_path: str,
+                    split: str,
```
Harold-lkk, line 121: whether need?

PR #1594, opened 2022-12-05
[Improve] support head loss or postprocessor is None for only infer

Harold-lkk:


mmocr/models/textdet/heads/base.py
```diff
@@ -53,15 +53,20 @@ class BaseTextDetHead(BaseModule):
...
+                 module_loss: Optional[Dict],
+                 postprocessor: Optional[Dict],
                  init_cfg: Optional[Union[Dict, List[Dict]]] = None) -> None:
         super().__init__(init_cfg=init_cfg)
-        assert isinstance(module_loss, dict)
-        assert isinstance(postprocessor, dict)
-
-        self.module_loss = MODELS.build(module_loss)
-        self.postprocessor = MODELS.build(postprocessor)
+        if module_loss is not None:
```
gaotongxiao, line 60:
```suggestion
        if getattr(self, 'module_loss', None) is not None:
```
Otherwise this line will result in undefined error when nothing is assigned to `module_loss`. Same for postprocessor

gaotongxiao (approved):

PR #1632, opened 2022-12-19
[Enhancement] Enhance FixInvalidPolygon, add RemoveIgnored transform

gaotongxiao:
Based on #1623.

## Motivation

1. There are a few more possible cases for invalid polygons which should be addressed in `FixInvalidPolygon`.
2. Some third-party augmentation libraries, such as ImgAug, may produce polygons not regarded as valid by `shapley`.
3. Sometimes there is a need to remove ignored instances from the pipeline. 

## Modifications

1. Enhance `FixInvalidPolygon` by tacking some corner cases which are usually seen in some datasets, and move it to `ocr_transforms.py` since it's general enough now.
2. Let `ImgAugWrapper` invokes `FixInvalidPolygon` to deal with some weird shape polygons by default.
3. Add `RemoveIgnored` transform to `ocr_transforms.py` as well.
4. Add `transform_utils.py` and implement `remove_pipeline_elements`.

### DIFF
--- a/mmocr/datasets/transforms/ocr_transforms.py
+++ b/mmocr/datasets/transforms/ocr_transforms.py
@@ -11,7 +11,9 @@ from mmcv.transforms.utils import avoid_cache_randomness, cache_randomness
 
 from mmocr.registry import TRANSFORMS
 from mmocr.utils import (bbox2poly, crop_polygon, is_poly_inside_rect,
-                         poly2bbox, remove_pipeline_elements, rescale_polygon)
+                         poly2bbox, poly2shapely, poly_make_valid,
+                         remove_pipeline_elements, rescale_polygon,
+                         shapely2poly)
 from .wrappers import ImgAugWrapper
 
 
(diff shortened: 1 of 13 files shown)


mmocr/datasets/transforms/ocr_transforms.py
```diff
@@ -613,3 +615,136 @@ def __repr__(self):
...
+
+    def transform(self, results: Dict) -> Dict:
+        remove_inds = np.where(results['gt_ignored'])[0]
+        if len(remove_inds) == len(results['gt_ignored']):
+            return None
+        return remove_pipeline_elements(results, remove_inds)
+
+
+@TRANSFORMS.register_module()
+class FixInvalidPolygon(BaseTransform):
```
Harold-lkk, line 649: It's maybe better to split into two `Transform`:
* RemovePolyPointLess: for removing the polygon whose length is less than `min_points`
* FixSelfCrossingOrSelfTouching(mode='fix/remove/ignore', remove_when_failed: bool, failed_replace_with_bbox: bool): for removing the polygon when it is self-crossing or self-torching

mmocr/datasets/transforms/wrappers.py
```diff
@@ -47,16 +47,24 @@ class ImgAugWrapper(BaseTransform):
...
             range [0.5, 3.0] for each side of images. Defaults to None.
+        fix_poly (bool): Whether to use
+            :class:`mmocr.datasets.transforms.FixInvalidPolygon` to
+            fix invalid polygons produced by ImgAug. Defaults to True.
     """
 
-    def __init__(self, args: Optional[List[Union[List, Dict]]] = None) -> None:
+    def __init__(self,
+                 args: Optional[List[Union[List, Dict]]] = None,
+                 fix_poly: bool = True) -> None:
```
Harold-lkk, line 57: fix_polygon_trans: dict
self.fix = Transform.build(fix_polygon)

mmocr/datasets/transforms/ocr_transforms.py
```diff
@@ -613,3 +615,136 @@ def __repr__(self):
...
+                        results['gt_polygons'][idx] = shapely2poly(
+                            poly_make_valid(shapely_polygon))
+                        # If an empty polygon is generated, it's still a bad
+                        # fix
+                        if len(results['gt_polygons'][idx]) == 0:
+                            raise ValueError
+                    # It's hard to fix, e.g. the "polygon" is a line or
+                    # a point
+                    except Exception:
+                        if 'gt_bboxes' in results:
```
Harold-lkk, line 732: add a flag `failed_replace_with_bbox` to decide whether replace polygon with bbox

mmocr/datasets/transforms/ocr_transforms.py
gaotongxiao: I don't think it's needed, because all of our transforms and evaluation metrics are built upon the assumption that these polygons are valid in `shapely`, and all this transform does is to ensure that no invalid polygons will be in the pipeline. Even though we have split this into two transforms, they will always be used together in order to not produce errors in the upcoming `shapely` operations.

In other words, if users don't need to rely on `shapely` in their implementation, they will never need to call `FixInvalidPolygon`. As for now, we can just keep this transform simple for most users.

mmocr/datasets/transforms/ocr_transforms.py
```diff
@@ -613,3 +615,144 @@ def __repr__(self):
...
+                    try:
+                        shapely_polygon = poly2shapely(polygon)
+                        if shapely_polygon.is_valid and len(
+                                polygon) >= self.min_poly_points * 2:
+                            continue
+                        results['gt_polygons'][idx] = shapely2poly(
+                            poly_make_valid(shapely_polygon))
+                        # If an empty polygon is generated, it's still a bad
+                        # fix
+                        if len(results['gt_polygons'][idx]) == 0:
```
Harold-lkk, line 733: should be removed?

Harold-lkk (approved):

PR #1635, opened 2022-12-20
[Checkpoints] Add ST-pretrained DB-series models and logs

gaotongxiao:
Also fixed the configs.


qiyueliuhuo:
When I set load_from = "this pretained DB models path"
then, I train `python tools/train.py configs/textdet/dbnet/dbnet_resnet18_fpnc_1200e_ctw1500.py`

but show 
The model and loaded state dict do not match exactly
image

Does this situation have any effect on model training?
I modified some keys of the model so that the parameters of some layers can be aligned and loaded, but there are still a lot of parameters that are not aligned

Harold-lkk (approved):

PR #1639, opened 2022-12-22
CodeCamp #116 Add SROIE to dataset preparer

FerryHuang:

## Motivation

CodeCamp-MMOCR-116

## Modification

Support SROIE dataset (text detection, text recognition, text spotting) on the DatasetPreparer

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

- [x] If the modification has potential influence on downstream or other related projects, this PR should be tested with some of those projects.
- [x] CLA has been signed and all committers have signed the CLA in this PR.


configs/textdet/_base_/datasets/sroie.py
```diff
@@ -0,0 +1,15 @@
+sroie_textdet_data_root = 'data/sroie'
```
gaotongxiao, line 1: This file can be deleted. `sroie` is not a prevailing dataset in OCR, and therefore we don't need to provide these base configs by default. Config generator will generate these base configs when needed.

configs/textrecog/_base_/datasets/sroie.py
```diff
@@ -0,0 +1,14 @@
+sroie_textrecog_data_root = 'data/sroie'
```
gaotongxiao, line 1: This file can be deleted for the same reason

dataset_zoo/sroie/textrecog.py
```diff
@@ -0,0 +1,77 @@
+data_root = 'data/sroie'
```
gaotongxiao, line 1: It seems this config is very similar to `textdet.py`, so we can take advantage of config inheritance like what you did in `textspotting.py`, or [totaltext's recog confg](https://github.com/open-mmlab/mmocr/blob/dev-1.x/dataset_zoo/totaltext/textrecog.py)

dataset_zoo/sroie/textspotting.py
```diff
@@ -0,0 +1,3 @@
+_base_ = ['textdet.py']
+
+data_converter = dict(type='TextSpottingDataConverter')
```
gaotongxiao, line 3: Missing one line:
```python
config_generator = dict(type='TextSpottingConfigGenerator')
```

mmocr/datasets/preparers/data_converter.py
```diff
@@ -164,10 +164,13 @@ def pair_gather(self, img_path: str, suffixes: List, rule: Sequence,
...
-            file2 = file.replace(osp.basename(file), file2)
-            file2 = file2.replace(self.img_dir, 'annotations')
-            files.append((file, file2))
+            if not re.match(rule[0], osp.basename(file)):
+                continue
+            else:
+                file2 = re.sub(rule[0], rule[1], osp.basename(file))
+                file2 = file.replace(osp.basename(file), file2)
+                file2 = file2.replace(self.img_dir, 'annotations')
+                files.append((file, file2))
```
gaotongxiao, line 173:
```suggestion
            file2 = re.sub(rule[0], rule[1], osp.basename(file))
            file2 = file.replace(osp.basename(file), file2)
            file2 = file2.replace(self.img_dir, 'annotations')
            files.append((file, file2))
```

dataset_zoo/sroie/sample_anno.md
```diff
@@ -0,0 +1,9 @@
+**Text Detection**
```
gaotongxiao, line 1:
```suggestion
**Text Detection, Text Recognition and Text Spotting**
```

mmocr/datasets/preparers/data_obtainer.py
```diff
@@ -148,6 +149,14 @@ def move(self, mapping: List[Tuple[str, str]]) -> None:
...
             dst = osp.join(self.data_root, dst)
+
+            if '*' in src.split('/')[-1] and not osp.exists(dst):
+                os.makedirs(dst)
+                for f in glob.glob(src):
+                    if osp.exists(f):
+                        shutil.move(f, dst)
+                continue
+
             if osp.exists(src) and not osp.exists(dst):
```
gaotongxiao, line 160: These lines can be optimized to be more general. The `if` condition is too strict and may only be applicable to this dataset. 

1. Wildcard (*) does not usually appear in the file path, and therefore we can use `glob` to match all the files whenever there is `*` in the file path.
2. `mmengine` has a util `mkdir_or_exist` to create the directory whenever needed.
3. Is line 156 really necessary? I think `glob.glob` should return a list of existing files.
4. Using continue to skip the next `if` block doesn't make the logic flow very clear. I'm in favor of `if...elif...` here

To conclude, the code can be rewritten as:
```python
import mmengine

#...

if '*' in src:
    mmengine.mkdir_or_exist(dst)
    for f in glob.glob(src):
        shutil.move(f, dst)
elif osp.exists(src) and not osp.exists(dst):
```

mmocr/datasets/preparers/parsers/sroie_parser.py
```diff
@@ -0,0 +1,72 @@
...
+from typing import List, Optional, Tuple
+
+from mmocr.utils import bbox2poly
+from ..data_preparer import DATA_PARSERS
+from .base import BaseParser
+
+
+@DATA_PARSERS.register_module()
+class SROIETextDetAnnParser(BaseParser):
+    """ICDAR Txt Format Text Detection Annotation Parser.
```
gaotongxiao, line 11:
```suggestion
    """SROIE Txt Format Text Detection Annotation Parser.
```

mmocr/datasets/preparers/parsers/sroie_parser.py
```diff
@@ -0,0 +1,72 @@
...
+        self.ignore = ignore
+        self.mode = mode
+        self.remove_strs = remove_strs
+        super().__init__(nproc=nproc)
+
+    def parse_file(self, file: Tuple, split: str) -> Tuple:
+        """Parse single annotation."""
+        img_file, txt_file = file
+        instances = list()
+        try:
```
gaotongxiao, line 53: Add a comment to tell others why we are catching the exception but doing nothing:

```suggestion
        # there might be some illegal symbols in the annotation
        # which cannot be parsed by loader
        try:
```

mmocr/datasets/preparers/data_obtainer.py
FerryHuang: Thanks for your suggestions and I have modified this code block. But with https://github.com/open-mmlab/mmocr/pull/1639#discussion_r1056922636 fixed, this reports error like 'path already exists' when running task-recog after task-det. Keeping the `not osp.exists(dst) `in `if` statement will be fine. So it's like this 
```
if '*' in src and not osp.exists(dst):
    mkdir_or_exist(dst)
    for f in glob.glob(src):
        shutil.move(f, dst)

elif osp.exists(src) and not osp.exists(dst):
```

gaotongxiao: It doesn't seem to address the key issue. Such an error occurs because `shutil` will check the existence of the destination file. If `dst` is just a directory path, `shutil` will do the check on the resulting filename, which is equivalent to `osp.exists(osp.join(dst, osp.basename(src))`.

Back to your implementation: `... and not osp.exists(dst)` only verifies the existence of the **directory** instead of a specific **file**. Suppose that we have placed an empty directory to `dst`, this implementation will still skip moving all the files inside. 

Therefore, it's better to just check the existence of each file with `osp.exists(osp.join(dst, osp.basename(f))` and only skip those existing ones.

mmocr/datasets/preparers/parsers/sroie_parser.py
```diff
@@ -0,0 +1,74 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from typing import List, Optional, Tuple
+
+from mmocr.utils import bbox2poly
+from ..data_preparer import DATA_PARSERS
+from .base import BaseParser
+
+
+@DATA_PARSERS.register_module()
+class SROIETextDetAnnParser(BaseParser):
```
Harold-lkk, line 10: what is the difference with ICDARTextDetAnnParaser

FerryHuang: The only difference is the try-except structure added to handle unicode errors that I want to skip. Should I merge this into the ICDARTextDetAnnParser?

gaotongxiao (commented):
It looks good overall, but does not pass our lint test.

image

To fix it, please install pre-commit hooks following

https://mmocr.readthedocs.io/en/dev-1.x/notes/contribution_guide.html#commit-your-changes

and then run `pre-commit run --all-files`, and it will help you fix the format issues.

gaotongxiao (approved):
LGTM. I've updated the downloading link for SROIE, since the google drive links doesn't work.

PR #1678, opened 2023-01-13
[Feature] Add RepeatAugSampler

gaotongxiao:


mmocr/datasets/samplers/repeat_aug.py
```diff
@@ -0,0 +1,100 @@
+import math
+from typing import Iterator, Optional, Sized
+
+import torch
+from mmcls.registry import DATA_SAMPLERS
```
Harold-lkk, line 5: mmocr

mmocr/datasets/samplers/repeat_aug.py
```diff
@@ -0,0 +1,100 @@
...
+            g.manual_seed(self.seed + self.epoch)
+            indices = torch.randperm(len(self.dataset), generator=g).tolist()
+        else:
+            indices = list(range(len(self.dataset)))
+
+        # produce repeats e.g. [0, 0, 0, 1, 1, 1, 2, 2, 2....]
+        indices = [x for x in indices for _ in range(self.num_repeats)]
+        # add extra samples to make it evenly divisible
+        padding_size = self.total_size - len(indices)
+        indices += indices[:padding_size]
```
Harold-lkk, line 76: indices = (indices *
                int(self.total_size / len(indices) + 1))[:self.total_size]
avoid world size >>  len(dataset)

Harold-lkk:
what is the difference with repeatdataset

Harold-lkk (approved):

PR #1702, opened 2023-02-02
[Enhancement] Dynamic return type for rescale_polygons

gaotongxiao:


mmocr/utils/polygon_utils.py
```diff
@@ -61,19 +62,22 @@ def rescale_polygons(polygons: Sequence[ArrayLike],
...
 
     Returns:
-        list[np.ndarray]: Rescaled polygons.
+        list[np.ndarray] or np.ndarray: Rescaled polygons. The type of the
+        return value depends on the type of the input polygons.
     """
     results = []
     for polygon in polygons:
         results.append(rescale_polygon(polygon, scale_factor, mode))
+    if isinstance(polygons, np.ndarray):
```
Harold-lkk, line 79: same for torch.tensor

Harold-lkk (approved):

PR #1749, opened 2023-02-27
[Fix] fix np.arange() output not stable error

KevinNuNu:

## Motivation
mmocr中pred_score_thrs生成的范围需要包含stop值，然而使用np.arange可能会因为浮点型数值精度问题导致不包含stop值。
image

https://numpy.org/doc/stable/reference/generated/numpy.arange.html
numpy官方也提示了这个潜在的问题是因为函数内部实现导致，并且建议如果要使用non-integer step最好还是使用numpy.linspace().

## Modification

使用`np.linspace(start, stop, int(np.round((stop - start) / step)) + 1, endpoint=True)`替换`np.arange(start, stop, step)`

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

- [ ] I have read and followed the workflow indicated in the [CONTRIBUTING.md](https://github.com/open-mmlab/mmocr/blob/main/.github/CONTRIBUTING.md) to create this PR.
- [ ] Pre-commit or linting tools indicated in [CONTRIBUTING.md](https://github.com/open-mmlab/mmocr/blob/main/.github/CONTRIBUTING.md) are used to fix the potential lint issues.
- [ ] Bug fixes are covered by unit tests, the case that causes the bug should be added in the unit tests.
- [ ] New functionalities are covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [ ] The documentation has been modified accordingly, including docstring or example tutorials.

**After PR**:

- [ ] If the modification has potential influence on downstream or other related projects, this PR should be tested with some of those projects.
- [ ] CLA has been signed and all committers have signed the CLA in this PR.

### DIFF
--- a/mmocr/evaluation/metrics/hmean_iou_metric.py
+++ b/mmocr/evaluation/metrics/hmean_iou_metric.py
@@ -44,7 +44,7 @@ class HmeanIOUMetric(BaseMetric):
         ignore_precision_thr (float): Precision threshold when prediction and\
             gt ignored polygons are matched. Defaults to 0.5.
         pred_score_thrs (dict): Best prediction score threshold searching
-            space. Defaults to dict(start=0.3, stop=0.9 step=0.1).
+            space. Defaults to dict(start=0.3, stop=0.9, step=0.1).
         strategy (str): Polygon matching strategy. Options are 'max_matching'
             and 'vanilla'. 'max_matching' refers to the optimum strategy that
             maximizes the number of matches. Vanilla strategy matches gt and


KevinNuNu:
但个人感觉以这种方式解决问题self.pred_score_thrs获取值得代码没有之前的优雅

mmocr/evaluation/metrics/hmean_iou_metric.py
```diff
@@ -63,14 +63,20 @@ class HmeanIOUMetric(BaseMetric):
...
         super().__init__(collect_device=collect_device, prefix=prefix)
         self.match_iou_thr = match_iou_thr
         self.ignore_precision_thr = ignore_precision_thr
-        self.pred_score_thrs = np.arange(**pred_score_thrs)
+        self.pred_score_thrs = np.linspace(
+            pred_score_thrs['start'],
+            pred_score_thrs['stop'],
+            int(
+                np.round((pred_score_thrs['stop'] - pred_score_thrs['start']) /
+                         pred_score_thrs['step'])),
```
gaotongxiao, line 78: How about `int((pred_score_thrs['stop'] - pred_score_thrs['start']) // pred_score_thrs['step'])`

gaotongxiao (commented):
This PR also changes the definition of `stop` a bit by excluding it from the range, but I don't think it's a good idea.

1. Breaking backward compatibility without sufficient reasons is not encouraged, since users might get a different result with the same config. And for now I don't see outstanding reasons for this change.
2. If we apply this change, we have to manually shift the stop point a bit if we want to filter the results with the original ending threshold. This becomes weirder when we want the threshold to stop at 1.

Therefore, it's better to roll back this definition change and stick to the old behavior.

mmocr/evaluation/metrics/hmean_iou_metric.py
KevinNuNu: It still can not include "stop" when using "start=0.3, stop=0.9, step=0.1"

PR #1824, opened 2023-03-30
update stitch_boxes_into_lines

cherryjm:

## Motivation

When there are boxes overlapped horizontally, the original `stitch_boxes_into_lines` function cannot properly stitch them.
Take `box_a` and `box_b` for example, if they are on the same line and `x1_a < x1_b < x2_a < x2_b`, they should be merged. Then it comes to the next box `box_c`, which is on the right. To determine whether it should be stitched, the correct `dist` should be `dist = left_boundary_of_box_c - right_boundary_of_merged_boxes`. But the current implementation calculates `dist` as `dist = left_boundary_of_box_c - right_boundary_of_box_b`. The problem is `right_boundary_of_merged_boxes != right_boundary_of_box_b` in this case.

## Modification

Modify the calculation of `dist` based on the above.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

- [ ] If the modification has potential influence on downstream or other related projects, this PR should be tested with some of those projects.
- [ ] CLA has been signed and all committers have signed the CLA in this PR.


gaotongxiao (approved):
Thanks! That's a good catch.

PR #2023, opened 2024-02-09
bumpy mmdet version

MiXaiLL76:
## Motivation

I want use mmocr with latest mmdet 

## Modification

Update mmdet dependence


MiXaiLL76:
@gaotongxiao

gaotongxiao (approved):

